PR #8, opened 2020-02-11
My solution

BohdanSalenik:
https://bohdansalenik.github.io/react_sum/


src/components/Sum.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+// eslint-disable-next-line
```
danheim: no way

PR #12, opened 2020-02-11
AnastasiiaBardyuzha: react_sum

AnastasiiaBardyuzha:
https://AnastasiiaBardyuzha.github.io/react_sum/

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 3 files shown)


src/components/Sum.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Sum(props) {
```
danheim:
```suggestion
function Sum({ a, b }) {
```

PR #17, opened 2020-02-11
solution

morfeii:
https://morfeii.github.io/react_sum/


src/components/Sum.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = (props) => {
+  const { a, b } = props;
```
irenhh: you can destructure this right in the props

PR #22, opened 2020-02-12
sum

cursorksu:
[DEMO LINK](https://cursorksu.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = (props) => {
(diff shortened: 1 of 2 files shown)


src/components/Sum.js
```diff
@@ -0,0 +1,26 @@
...
+};
+
+Sum.propTypes = {
+  data: PropTypes.shape({
+    a: PropTypes.number.isRequired,
+    b: PropTypes.number.isRequired,
+  }),
+};
+
+Sum.defaultProps = {
```
irenhh: if you have defaultProps, you don't need to mention `isRequired`

PR #25, opened 2020-02-12
Serebrenykov: Develop

Alex-8-8:
[DEMO LINK](https://Alex-8-8.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,13 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function Sum(props) {
+  return (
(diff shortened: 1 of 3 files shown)


src/components/Sum.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function Sum(props) {
```
irenhh: use desctructuring

irenhh: remove `default`

irenhh: since it's name export you don't need default here

PR #26, opened 2020-02-12
solution

oksanaDaniluk:
https://github.com/mate-academy/react_task-guideline#react-tasks-guideline

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Sum({ a, b }) {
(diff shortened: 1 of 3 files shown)


src/components/Sum.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Sum({ a, b }) {
```
gestap: так коротше
```
const Sum = ({ a, b }) => <p>{...}</p>
```

oksanaDaniluk: done

PR #27, opened 2020-02-12
add solution

Yuliya312:
[DEMO LINK](https://Yuliya312.github.io/react_sum/)


src/components/Sum.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Sum(props) {
```
irenhh: you can destructure right in the args

PR #53, opened 2020-04-22
Solution

TetianaZakharova:
[DEMO LINK](https://TetianaZakharova.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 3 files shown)


src/components/Sum.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = (props) => {
```
vkryvytskyy, line 4:
```suggestion
const Sum = ({ a, b }) => {
```

PR #86, opened 2020-06-21
Solution

ogsevko:
[DEMO LINK](https://ogsevko.github.io/react_sum/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,8 +1,26 @@
 import React from 'react';
 import './App.css';
+import PropTypes from 'prop-types';
 
-const App = () => (
-  <h1>React sum</h1>
+const App = ({ a, b }) => (
+  <h1>
+    Sum of
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
+  </h1>
 );
+
+App.propTypes = {
+  a: PropTypes.number.isRequired,
+  b: PropTypes.number.isRequired,
+};
 
 export default App;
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -1,8 +1,26 @@
 import React from 'react';
 import './App.css';
+import PropTypes from 'prop-types';
 
-const App = () => (
-  <h1>React sum</h1>
+const App = ({ a, b }) => (
```
solaryasha, line 5:
```suggestion
const App = props => (
```
In React, you pass some data to component with `props`

src/App.js
```diff
@@ -1,8 +1,26 @@
...
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
```
solaryasha, line 17:
```suggestion
{
    `Sum of ${props.a} and ${props.b} is ${props.a + props.b}`
}
```

ogsevko: Yeah, I tried to do this at first, but eslint doesn't think that it's OK to import React and then not use it, so it throws an error.

src/App.js
ogsevko: I am well aware of that :) But it's destructuring assignment and as far as I know it's ok to use it. Also, when you have to use some property multiple times it's just more convenient to write `a` rather than `props.a`

solaryasha: Alright, maybe you are right 👍

danheim: Yaroslav, you are right. Greate job.

src/index.js
```diff
@@ -3,6 +3,10 @@ import ReactDOM from 'react-dom';
 import App from './App';
 
 ReactDOM.render(
-  <App />,
+  <React.Fragment>
```
danheim, line 6: use short form
```suggestion
  <>
```

PR #101, opened 2020-06-23
Develop

coroboX:
[DEMO LINK](https://coroboX.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,12 @@
+import React from 'react';
+
+const Sum = (props) => {
+  // eslint-disable-next-line no-console
+  console.log(props);
+
+  return (
+    <p>Sum of {props.a} and {props.b} is {+props.a + props.b}</p>
+  );
+};
+
+export default Sum;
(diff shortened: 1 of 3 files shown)


src/components/Sum.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const Sum = (props) => {
```
solaryasha, line 3: It's better to use object destructuring here
```suggestion
const Sum = ({a, b}) => {
```

src/components/Sum.js
```diff
@@ -0,0 +1,12 @@
...
+  // eslint-disable-next-line no-console
+  console.log(props);
```
danheim, line 5: remove

src/components/Sum.js
danheim, line 3: деструктурируй пропсы

src/components/Sum.js
```diff
@@ -0,0 +1,12 @@
...
+
+  return (
+    <p>Sum of {props.a} and {props.b} is {+props.a + props.b}</p>
+  );
```
danheim, line 9: return не нужно

PR #102, opened 2020-06-23
Sum

tyooma:
[DEMO LINK](https://tyooma.github.io/react_sum/

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = props => (
+  <p>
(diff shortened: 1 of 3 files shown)


src/components/Sum.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = props => (
+  <p>
+    {`Sum of ${props.a} and ${props.b} is ${+props.a + +props.b}`}
+  </p>
+);
+
+export { Sum };
```
danheim, line 10:
```suggestion
export const Sum = props => (
  <p>
    {`Sum of ${props.a} and ${props.b} is ${+props.a + +props.b}`}
  </p>
);
```

PR #142, opened 2020-08-31
Solution

martakupka:
[DEMO LINK](https://martakupka.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = props => (
+  <p className="sum">
+    Sum of
+    {' '}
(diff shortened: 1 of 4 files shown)


src/components/Sum.js
```diff
@@ -0,0 +1,25 @@
...
+    {props.a}
+    {' '}
+    and
+    {' '}
+    {props.b}
+    {' '}
+    is
+    {' '}
+    {props.a + props.b}
+  </p>
```
leonbohdan, line 17: here you can use like this
```{`Sum of ${a} and ${b} is ${a + b}`}```

leonbohdan (approved):
Good work

vsm2k17 (approved):

PR #228, opened 2021-01-11
added solution

MarchD:
[DEMO LINK](https://MarchD.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,13 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({ a = 1, b = 2 }) => (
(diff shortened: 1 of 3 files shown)


troospieler (approved):
all good

src/components/Sum.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({ a = 1, b = 2 }) => (
```
aionova, line 4: If either give default values to your props, or indicate them as required in PropTypes.

aionova (changes requested):
Good start!
You can't indicate props as required and give them default values at the same time. If they are required, default value is redundant.

aionova (approved):
Awesome!

PR #233, opened 2021-01-11
solution

dima-bond:
[DEMO LINK](https://dima-bond.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({ a, b }) => (
+  <p>
+    Sum of
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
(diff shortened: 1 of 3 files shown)


chri911 (approved):
So beautiful)

src/components/Sum.js
```diff
@@ -0,0 +1,25 @@
...
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
```
aionova, line 16:
```suggestion
    {`Sum of ${a} and ${b} is ${a + b}`}
```
Make it easy to read

aionova (changes requested):
Good start!

src/components/Sum.js
dima-bond: thanks, fixed

aionova (approved):
Great!

PR #271, opened 2021-02-28
Props task React Sum

MrKartoshka:
Create Sum to calculate a + b sum and display result 
Created Sum folder to contain Sum component

### DIFF
--- a/src/components/Sum/Sum.js
+++ b/src/components/Sum/Sum.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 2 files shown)


src/components/Sum/Sum.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Sum(props) {
```
vpolets, line 4:
```suggestion
function Sum({ a, b }) {
```

vpolets (approved):

PR #345, opened 2021-05-17
React Sum

nazarmatsevych:


provolovskyi-illia (approved):
Молодец)

src/App.js
```diff
@@ -1,8 +1,26 @@
...
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
```
VinnieJ-2k20, line 17: There is much easier way to render this text:
```suggestion
    Sum of
    {`Sum of ${a} and ${b} is ${a + b}`}
```

VinnieJ-2k20 (changes requested):
You need to create `Sum` component and render it three times inside `App`.

src/App.js
```diff
@@ -1,8 +1,16 @@
 import React from 'react';
 import './App.scss';
+import PropTypes from 'prop-types';
 
-const App = () => (
-  <h1>React sum</h1>
+const App = ({ a, b }) => (
+  <h1>
+    {`Sum of ${a} and ${b} is ${a + b}`}
+  </h1>
```
VinnieJ-2k20, line 8: Inside `App` component there should be three instances of `Sum` components (that you need to create).

VinnieJ-2k20 (changes requested):
Some more work needed. Details in comments.

VinnieJ-2k20 (approved):
Good job 👍

PR #376, opened 2021-05-18
The Sum

leonid-vegera:
[DEMO LINK](https://leonid-vegera.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.js
+++ b/src/components/Sum.js
@@ -1,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({ a, b }) => (
+  <p className="Numbers">
+    {`Sum of ${a} and ${b} is ${a + b}`}
+  </p>
+);
+
+Sum.defaultProps = {
+  b: 0,
+};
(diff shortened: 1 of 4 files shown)


SergeyZub (approved):

src/components/Sum.js
```diff
@@ -0,0 +1,19 @@
...
+);
+
+Sum.defaultProps = {
+  b: 0,
+};
+
+Sum.propTypes = {
+  a: PropTypes.number.isRequired,
+  b: PropTypes.number,
+};
```
VinnieJ-2k20, line 17: Don't clearly understand why `a` is required and `b` is not. But OK.

VinnieJ-2k20 (approved):
Good job 👍 

Added one comment on props logic.

src/components/Sum.js
leonid-vegera: > Don't clearly understand why `a` is required and `b` is not. But OK.

Such a decision didn't have any sense, just practiced. For self-affirmation)

PR #388, opened 2021-05-18
React sum

Allrightman:
[DEMO LINK](https://allrightman.github.io/react_sum/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,8 +1,27 @@
 import React from 'react';
 import './App.scss';
+// eslint-disable-next-line import/extensions
+import Sum from './components/Sum.jsx';
+
+const first = {
+  a: 2, b: 2,
+};
+
+const second = {
+  a: 10, b: 0,
+};
+
+const third = {
+  a: -5, b: 5,
+};
 
 const App = () => (
-  <h1>React sum</h1>
+  <div className="App">
+    <h1>React sum</h1>
+    <Sum a={first.a} b={first.b} />
+    <Sum a={second.a} b={second.b} />
+    <Sum a={third.a} b={third.b} />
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+// eslint-disable-next-line react/prop-types
```
aliceblunt8, line 3: don't disable linter in this case, instead add prop types to this component

src/components/Sum.jsx
```diff
@@ -0,0 +1,14 @@
...
+const Sum = ({ a, b }) => {
+  const text = `Sum of ${a} and ${b} is ${a + b}`;
```
aliceblunt8, line 5: no need to create a separate variable for this, just add to jsx in return statement

src/App.js
```diff
@@ -1,8 +1,27 @@
...
+  a: 2, b: 2,
+};
+
+const second = {
+  a: 10, b: 0,
+};
+
+const third = {
+  a: -5, b: 5,
+};
```
aliceblunt8, line 16: remove these, pass just numbers

src/App.js
```diff
@@ -1,8 +1,27 @@
 import React from 'react';
 import './App.scss';
+// eslint-disable-next-line import/extensions
+import Sum from './components/Sum.jsx';
```
aliceblunt8, line 4: remove extension and linter will be ok
```suggestion
import Sum from './components/Sum';
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #465, opened 2021-08-27
solution react-sum done

AndreySerhovetc:
[DEMO LINK](https://AndreySerhovetc.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -4,12 +4,7 @@ import PropTypes from 'prop-types';
 
 const Sum = ({ a, b }) => (
   <p>
-    {'Sum of '}
-    {a}
-    {' and '}
-    {b}
-    {' is '}
-    {a + b}
+    {`Sum of ${a} and ${b} is ${a + b}`}
   </p>
 );
 


src/components/Sum.jsx
```diff
@@ -1 +1,20 @@
...
+    {'Sum of '}
+    {a}
+    {' and '}
+    {b}
+    {' is '}
+    {a + b}
+  </p>
+);
+
+Sum.defaultProps = {
```
sasha-morozov, line 15: Apart from default values, you also need to check the data type of props.

sasha-morozov (approved):
Good job! Left one suggestion above

src/components/Sum.jsx
AndreySerhovetc: ok)

src/components/Sum.jsx
```diff
@@ -1 +1,25 @@
...
+import PropTypes from 'prop-types';
+
+const Sum = ({ a, b }) => (
+  <p>
+    {'Sum of '}
+    {a}
+    {' and '}
+    {b}
+    {' is '}
+    {a + b}
```
varseniuk, line 12:
```suggestion
    {`Sum of ${a} and ${b} is ...${a + b}`}
```

varseniuk (approved):
👍

src/components/Sum.jsx
AndreySerhovetc: fixed

PR #466, opened 2021-08-28
sum react

DmytroMykoliv:
[DEMO LINK](https://dmytromykoliv.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,28 @@
 // Don't forget to add default values an check types of props
+import React from 'react';
+import PropTypes from 'prop-types';
+import './Sum.scss';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
+  <p className="addition">
+    {'Sum of '}
+    {a}
+    {' and '}
+    {b}
+    {' is '}
+    {a + b}
+  </p>
+);
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
+
+Sum.propTypes = {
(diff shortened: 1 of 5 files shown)


illiarerun (approved):
Good job 🔥 I really like that you've worked with styling aswell

src/components/Sum.scss
```diff
@@ -0,0 +1,11 @@
+.addition {
```
varseniuk, line 1:
```suggestion
.sum {
```
It's a good practice to make the name of the main class of the component should be the same as the filename

src/components/Sum.jsx
```diff
@@ -1 +1,27 @@
...
+  a,
+  b,
+}) => (
+  <p className="addition">
+    {'Sum of '}
+    {a}
+    {' and '}
+    {b}
+    {' is '}
+    {a + b}
```
varseniuk, line 15:
```suggestion
    {`Sum of ${a} and ${b} is ...${a + b}`}
```
Actually it could be written like this

src/components/Sum.jsx
```diff
@@ -1 +1,27 @@
-// Don't forget to add default values an check types of props
+import React from 'react';
+import PropTypes from 'prop-types';
+import './Sum.scss';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
```
```suggestion
export const Sum = ({ a,  b }) => (
```
No need to split it to multiple lines if there are less than 3 arguments

varseniuk (approved):
Good job. Left some cooments. Pay attention

PR #468, opened 2021-08-29
Added Sum component with default props and propTypes

ihor-jpeg:
- [DEMO LINK](https://ihor-jpeg.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,27 @@
-// Don't forget to add default values an check types of props
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
+  <p>
+    {'Sum of '}
+    {a}
+    {' and '}
+    {b}
+    {' is '}
+    {a + b}
+  </p>
+);
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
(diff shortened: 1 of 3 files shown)


TarasHoliuk (approved):
everything alright!

src/components/Sum.jsx
```diff
@@ -1 +1,27 @@
-// Don't forget to add default values an check types of props
+import React from 'react';
+
+import PropTypes from 'prop-types';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
```
```suggestion
export const Sum = ({ a, b }) => (
```
When there are less than 3 arguments you can write them in one line

src/components/Sum.jsx
```diff
@@ -1 +1,27 @@
...
+  <p>
+    {'Sum of '}
+    {a}
+    {' and '}
+    {b}
+    {' is '}
+    {a + b}
```
varseniuk, line 15:
```suggestion
    {`Sum of ${a} and ${b} is ...${a + b}`}
```
Actually it could be written like this

varseniuk (approved):
👍

src/components/Sum.jsx
ihor-jpeg: hahah indeed, how did i miss that. thank you!

PR #478, opened 2021-08-31
Develop

Vova-Stoliar:
[DEMO LINK](https://Vova-Stoliar.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,19 @@
 // Don't forget to add default values an check types of props
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    {'Sum of '}
+    {a}
+    {' and '}
+    {b}
+    {' is '}
+    {a + b}
+  </p>
+);
+
+Sum.propTypes = {
+  a: PropTypes.number.isRequired,
+  b: PropTypes.number.isRequired,
+};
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,19 @@
...
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    {'Sum of '}
+    {a}
+    {' and '}
+    {b}
+    {' is '}
+    {a + b}
```
AndreySerhovetc, line 12: You can use this {`Sum of ${a} and ${b} is ${a + b}`}

AndreySerhovetc (approved):

src/components/Sum.jsx
varseniuk: Yep. It would be more readable

src/components/Sum.jsx
```diff
@@ -1 +1,19 @@
 // Don't forget to add default values an check types of props
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
varseniuk, line 5:
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```
You could use defaultProps for this instead

varseniuk (approved):
Good 👍 
But pay attention to my comments

PR #484, opened 2021-08-31
first try

zadorojniiarsen:


RomSmile:
Good solution, but tou have a one mistake.

src/App.jsx
```diff
@@ -1,15 +1,24 @@
...
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum
+      x={2}
+      y={3}
+    />
```
varseniuk, line 10:
```suggestion
    <Sum x={2} y={3} />
```
You can write it in one line. There are not so many props here

src/components/Sum.jsx
```diff
@@ -1 +1,25 @@
...
+
+const Sum = ({ x, y }) => (
+  <div>
+    {'Sum of '}
+    {x}
+    {' and '}
+    {y}
+    {' is '}
+    {x + y}
+  </div>
```
varseniuk, line 12:
```suggestion
  <div>
    Sum of {x} and {y} is {x + y}
  </div>
```
As a variant

varseniuk (approved):
👍

PR #485, opened 2021-08-31
task is done

Oleksandr-Zhitenko:
[DEMO LINK](https://Oleksandr-Zhitneko.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,15 +1,24 @@
 import React from 'react';
 import './App.scss';
+import Sum from './components/Sum';
 
-const App = () => (
+const App = (a, b) => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum 
+      a={2}
+      b={3}
+    />
+    <Sum 
+      a={-5}
+      b={5}
+    />
+    <Sum 
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
+    <Sum/>
   </>
 );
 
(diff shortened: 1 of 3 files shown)


RomSmile:
Good solution, but you have one mistake.

zadorojniiarsen (approved):

src/App.jsx
```diff
@@ -1,15 +1,24 @@
 import React from 'react';
 import './App.scss';
+import Sum from './components/Sum';
 
-const App = () => (
+const App = (a, b) => (
```
varseniuk, line 5:
```suggestion
const App = () => (
```
Why do you need these arguments here? You don't use them anywhere

src/App.jsx
```diff
@@ -1,15 +1,24 @@
...
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum 
+      a={2}
+      b={3}
+    />
```
varseniuk, line 10:
```suggestion
    <Sum a={2} b={3} />
```
There are not so many props here so you may write them in one line

src/components/Sum.jsx
```diff
@@ -1 +1,31 @@
 // Don't forget to add default values an check types of props
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({
+  a,
+  b,
+}) => {
```
```suggestion
const Sum = ({ a, b }) => {
```

src/components/Sum.jsx
```diff
@@ -1 +1,31 @@
...
+  return (
+    <p>
+      {`Sum of `}
+      {a}
+      {' and '}
+      {b}
+      {' is '}
+      {a + b}
```
varseniuk, line 16:
```suggestion
      Sum of {a} and {b} is {a + b}
```
As a variant

varseniuk (changes requested):
Let's remove unnecessary arguments from your `App` component

varseniuk (approved):

PR #579, opened 2021-11-02
Solution

MariiaObraztsova:


cyb1ove (approved):

src/components/Sum.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
```
VinnieJ-2k20, line 7: Try not to add unnecessary line-breaks.
```suggestion
export const Sum = ({ a, b }) => (
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #588, opened 2021-11-03
React SUM

HaqpaHR:
-[DEMO](https://haqpahr.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -6,13 +6,7 @@ function Sum({ a = 0, b = 0 }) {
 
   return (
     <p>
-      {`Sum of
-      ${a}
-      and
-      ${b}
-      is
-      ${sum}
-    `}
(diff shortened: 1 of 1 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <p>
+      {`Sum of
+      ${a}
+      and
+      ${b}
+      is
+      ${sum}
+    `}
+    </p>
```
DammneD, line 16: I think it looks awful. It's a short one, so maybe write it in one string?)

DammneD (approved):
Everything is fine, but this string... Leaving it like this will be on your conscience 😞

src/components/Sum.jsx
HaqpaHR: solve it, I just saw it in a hint of the readme.

VinnieJ-2k20 (approved):
Good job 👍

PR #611, opened 2021-12-02
some sort of solution

JamesSunderlanddd:
[DEMO LINK](https://jamessunderlanddd.github.io/react_sum/)

### DIFF
--- a/src/components/Sum/Sum.jsx
+++ b/src/components/Sum/Sum.jsx
@@ -1,0 +1,9 @@
+import React from 'react';
+
+export function Sum({ num1 = 0, num2 = 0 }) {
+  return (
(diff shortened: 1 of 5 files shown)


Vitaliy-0 (approved):
In future, don't add README file, and also you can add library PropsType and PropsDefault.
Good Job!

src/components/Sum/Sum.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export function Sum({ num1 = 0, num2 = 0 }) {
```
aionova, line 3: you can use arrow function if you don't write anything before return

aionova (approved):

PR #614, opened 2021-12-03
completed react sum

i-stanko:


SerhiiKirik (approved):
Well done 🚀 

Try to use commands to add and commit
`git add src/(...file)` for add
`git commit -m 'some message'` for commit without extra file

aionova (approved):
GJ!

PR #620, opened 2021-12-03
some message

dlugash:
[DEMO LINK](https://dlugash.github.io/react_sum/)


src/components/Sum.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const Sum = props => (
+  <p className="Sum">
+    {`Sum of ${props.x} and ${props.y} is ${props.x + props.y}`}
+  </p>
```
volodymyr-soltys97, line 6: Try to use props destructuring in this task

volodymyr-soltys97 (changes requested):
Very interesting way 👍

src/App.scss
```diff
@@ -1 +1,4 @@
 // styles go here
+.Sum {
```
AlinaMatuschak, line 2: create new style file for Sum component

AlinaMatuschak (approved):

PR #627, opened 2021-12-03
implemented task

tmrsl:
[DEMO LINK](https://tmrsl.github.io/react_sum/)


src/components/Sum.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+function Sum({ a, b }) {
+  const sum = a + b;
+
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${sum}`}
+    </p>
+  );
+}
```
kurtkobein, line 12: Do you need a sum variable?

kurtkobein (approved):
👍

src/components/Sum.jsx
illiarerun: @kurtkobein It's actually better to move all the calculation and other logic from jsx code to make it more readable. So, yeah, using variable for sum is good here

illiarerun (approved):
Great job 🔥 But remember that you shouldn't add unnecessary files to your commits (package-lock.json for example)

PR #635, opened 2021-12-05
First solution

WorkAccCom:


src/components/Sum.jsx
```diff
@@ -0,0 +1,32 @@
...
+      {' '}
+      {a}
+      {' '}
+      and
+      {' '}
+      {b}
+      {' '}
+      is
+      {' '}
+      {sum}
```
DudarDaryna, line 19: It can be shorter ;)
```suggestion
      {`Sum of ${a} and ${b} is ${sum}`}
```

DudarDaryna (approved):
GJ

src/components/Sum.jsx
illiarerun: Agree with Daryna, you should fix it, it's really hard to read

illiarerun (changes requested):
Good job 🔥 Fix Daryna's comment now:

illiarerun (approved):
GJ 🔥

PR #642, opened 2021-12-06
solution

sviatoslav-myhal:
https://sviatoslav-myhal.github.io/react_sum/

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
+  <p>{`Sum of ${a} and ${b} is ${+a + +b}`}</p>
+);
+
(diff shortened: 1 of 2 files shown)


aldmpr (approved):
Blow mind job!

src/components/Sum.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
+  <p>{`Sum of ${a} and ${b} is ${+a + +b}`}</p>
```
TatianaMorozova04, line 5: You can pass proops not as a string, but as a JS expression. Then you will not need to use {+ a + + b}

TatianaMorozova04 (approved):

AlinaMatuschak (changes requested):
fix @TatianaMorozova04 comment

PR #672, opened 2022-01-13
Develop

OlegShanin:
[DEMO LINK](https://OlegShanin.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({
+  a,
+  b,
+}) => (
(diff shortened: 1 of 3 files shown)


Sesile26 (changes requested):

Sesile26 (approved):

src/components/Sum.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({
+  a,
+  b,
+}) => (
```
Olexandr6, line 7: It looks better
```suggestion
const Sum = ({ a, b }) => (
```

Olexandr6 (changes requested):
Some code styles should be fixed

Olexandr6 (approved):

PR #674, opened 2022-01-13
Sum of numbers

Toros-Zz:
# Sum component
- [DEMO LINK](https://Toros-Zz.github.io/react_sum/)


Sesile26 (approved):

src/components/Sum.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({ a, b }) => (
+  <>
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  </>
```
Olexandr6, line 7: You don't need to use `React.Fragment` here, cause you have only one child element - `<p>...</p>`. React.Fragment you should use in case you need to cover 2 or more elements in parent cover.
```suggestion
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```

Olexandr6 (changes requested):
Check comment

Olexandr6 (approved):

PR #700, opened 2022-01-14
Develop

ErnestoFernandezUA:
[DEMO LINK](https://ErnestoFernandezUA.github.io/react_sum/)


src/components/Sum.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Sum({ a, b }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${+a + +b}`}</p>
+  );
```
Elena-brv, line 7: - use arrow function and not use return here
- why do you need to convert `a` and `b` to numbers? they are already numbers

Elena-brv (changes requested):

Elena-brv (approved):

PR #709, opened 2022-02-08
Solution_0.2

andriy-fesych:
[DEMO LINK](https://andriy-fesych.github.io/react_sum/)

Variation where component Sum takes props with destructuring for a and b executed in first commit.
I intentionally used 'props' and called variables a and b via props.a / props.b relatively.


src/components/Sum.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import './Sum.scss';
+import PropTypes from 'prop-types';
+
+// eslint-disable-next-line arrow-parens
```
danylo-onishchuk, line 5: why you need this rule?

danylo-onishchuk (changes requested):

src/components/Sum.jsx
andriy-fesych: because I want to use single function argument :disappointed:

danylo-onishchuk (approved):

PR #729, opened 2022-02-11
Solution

Vonatlus:
- [DEMO LINK](http://vonatlus.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -8,6 +8,7 @@ const App = () => (
     <Sum a={5} b={-5} />
     <Sum a={10} />
     <Sum b={5} />
+    <Sum />
   </>
 );
 


vsuslov29 (approved):
Гуд жоб!

src/App.jsx
```diff
@@ -9,6 +10,28 @@ const App = () => (
...
+        a: 2,
+        b: 3,
+      },
+    )}
+    {Sum(
+      {
+        a: 5,
+        b: -5,
+      },
+    )}
```
Anna-Tkachenko, line 24: Check in video how you should add components and pass needed arguments

Anna-Tkachenko: <Sum a=... b=... />

Anna-Tkachenko (changes requested):

PR #745, opened 2022-02-23
react_sum

IvanShulhan:
[DEMO LINK](https://ivanshulhan.github.io/react_sum/)


src/components/Sum.jsx
```diff
@@ -0,0 +1,23 @@
...
+import PropTypes from 'prop-types';
+
+const resultColor = {
+  color: '#b83c3c',
+};
+
+export const Sum = ({ a, b }) => (
+  <p>
+    {`Sum of ${a} and ${b} is `}
+    <span style={resultColor}>{a + b}</span>
```
vitaliy-martynyuk, line 11: it is better to avoid using inline styles. better move it to a separate class

vitaliy-martynyuk (changes requested):

vitaliy-martynyuk (approved):
GJ!

PR #750, opened 2022-03-08
Implement Sum component

Andr1yK:
[DEMO LINK](https://Andr1yK.github.io/react_sum)


src/components/Sum.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
```
Ihor-Karpyn, line 7:
```suggestion
export const Sum = ({ a, b }) => (
```

PR #768, opened 2022-03-23
Develop

RomaBoreiko:
[DEMO LINK](https://RomaBoreiko.github.io/react_sum/)


src/components/Sum.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const Sum = ({ a = 0, b = 0 }) => (
```
krolich500grn, line 3: use defaultprops here

krolich500grn (changes requested):

Ihor-Karpyn (approved):

PR #816, opened 2022-04-29
Solution

AmbeMO:


src/components/Sum.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = props => (
```
Yurii837, line 4: If you used named export "export function Sum" no need to change test file.

Yurii837 (approved):
Nice solution.

src/components/Sum.jsx
MaksimusCRS, line 4:
```suggestion
const Sum = {a, b} => (
```

src/components/Sum.jsx
```diff
@@ -0,0 +1,30 @@
...
+    {' '}
+    {props.a}
+    {' '}
+    and
+    {' '}
+    {props.b}
+    {' '}
+    is
+    {' '}
+    {props.a + props.b}
```
MaksimusCRS, line 16:
```suggestion
    {`Sum of ${a} and ${b} is ${a+b}`}
```

MaksimusCRS (approved):

PR #835, opened 2022-05-24
solution

vlad-demchuk:


src/components/Sum.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const Sum = ({ a = 0, b = 0 }) => (
+  <d>
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  </d>
+);
```
khozhainov-aleksandr, line 9: How does Sum do the calculations??

khozhainov-aleksandr (changes requested):
No preview link 🧐 and Good job ))

src/components/Sum.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const Sum = ({ a = 0, b = 0 }) => (
+  <d>
```
khozhainov-aleksandr, line 4: <d> ??? why is this?

khozhainov-aleksandr (changes requested):

src/components/Sum.jsx
vlad-demchuk: fixed)

khozhainov-aleksandr:
404 There isn't a GitHub Pages site here.

try to do `npm run deploy` 😉

Denchepornyuk (approved):

PR #841, opened 2022-05-24
solutionI

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,26 @@
+import React from 'react';
+
+function Sum({ a, b }) {
+  return (
+    <p>
+      Sum of
+      {' '}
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,26 @@
...
+      {' '}
+      {a}
+      {' '}
+      and
+      {' '}
+      {b}
+      {' '}
+      is
+      {' '}
+      {Number(a) + Number(b)}
```
vlad-demchuk, line 16: you should use interpolation  here) 
[CODE STYLE] - Use string interpolation inside tag content(checklist)

Denchepornyuk (approved):

PR #842, opened 2022-05-24
First try

RomekIvantsiv:


alexandra-protyanova (approved):
🌟☄️

src/components/Sum.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const Sum = ({ a = 0, b = 0 }) => (
```

Denchepornyuk (changes requested):

PR #844, opened 2022-05-25
solution

alexandra-protyanova:


vlad-demchuk (approved):

src/components/Sum.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #848, opened 2022-05-25
react_sum

anastasiia-nikita:
[DEMO LINK](https://anastasiia-nikita.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,10 +1,12 @@
 import React from 'react';
 
-export const Sum = ({
-  a = 0,
-  b = 0,
-}) => (
+export const Sum = ({ a, b }) => (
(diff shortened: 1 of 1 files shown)


vlad-demchuk (approved):
looks beautiful) you could use defaultProps to set default properties)

src/components/Sum.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
Denchepornyuk, line 6: add propTypes for the component

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #849, opened 2022-05-25
Develop

yelnikov-andrii:
[DEMO LINK](https://yelnikov-andrii.github.io/react_sum/)


src/components/Sum.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const Sum = ({ a = 0, b = 0 }) => (
```

Denchepornyuk (changes requested):

src/components/Sum.jsx
```diff
@@ -0,0 +1,18 @@
...
+  a: 0,
+  b: 0,
+};
+
+Sum.propTypes = {
+  a: PropTypes.number,
+  b: PropTypes.number,
+};
+
+export default Sum;
```
Anna-Tkachenko, line 18: use named export to fix the error

src/components/Sum.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({ a, b }) => (
```
Anna-Tkachenko, line 4:
```suggestion
export const Sum = ({ a, b }) => (
```

Anna-Tkachenko (approved):

PR #858, opened 2022-05-26
react sum

valerii-derkach:


src/components/Sum.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+);
```
iwanttosmile4u, line 5: Add propTypes for this component =)

src/components/Sum.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
iwanttosmile4u, line 3: Also you can write default values using defaultProps

iwanttosmile4u (commented):
Good! Just have a little ramarks)

Denchepornyuk (approved):

PR #859, opened 2022-05-26
task is solved

enrikadej:


valerii-derkach (changes requested):
empty file Sum.jsx
I can`t see your code

Anna-Tkachenko (approved):
⭐️

PR #860, opened 2022-05-26
First_commit

igoree88:
[DEMO LINK](https://igoree88.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,16 @@
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
+  // eslint-disable-next-line react/react-in-jsx-scope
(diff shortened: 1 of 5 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,16 @@
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
+  // eslint-disable-next-line react/react-in-jsx-scope
```
Anna-Tkachenko, line 4: You should add React import in this file and delete this line

Anna-Tkachenko (changes requested):

src/components/Sum.jsx
igoree88: Fixed

Anna-Tkachenko (approved):
⭐️

PR #868, opened 2022-05-29
solution

LukaszGryziewicz:
[DEMO LINK](https://LukaszGryziewicz.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
+  <p>
+    {`Sum of ${a} and ${b} is ${a + b}`}
+  </p>
+);
+
+Sum.defaultProps = {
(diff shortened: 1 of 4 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
```
efdalyalcin, line 4: I think you could directly give default values here like `{ a = 0, b = 0 }` but your way still works.

efdalyalcin (approved):
Well done 🔥

src/components/index.jsx
```diff
@@ -0,0 +1 @@
+export * from './Sum';
```
VinnieJ-2k20, line 1: This file should be `.js`, it has no JSX syntax

VinnieJ-2k20 (approved):
Good job 👍

PR #870, opened 2022-05-29
Task complete

rHashTag:
[DEMO LINK](https://rHashTag.github.io/react_sum/)


59MrRobot (approved):
Good work,

I'm not sure if we were meant to use proptypes or not for this task so I'll send you through for a mentor review and see what they suggest

VinnieJ-2k20 (approved):
Good job 👍 

PropTypes is optional here, we switch to TypeScript next anyway.

PR #871, opened 2022-05-30
solution

Uktam19980416:
[DEMO LINK](https://Uktam19980416.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => {
+  const res = a + b;
+
+  return (
+    <p>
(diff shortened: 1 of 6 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,29 @@
...
+    <p>
+      Sum of
+      { a }
+      {' '}
+      and
+      {b}
+      {' '}
+      is
+      { res }
+    </p>
```
59MrRobot, line 17: This seems a bit difficult to read, why not use interpolation instead?

59MrRobot (changes requested):
Good work, just one suggestion

59MrRobot (approved):
👍

VinnieJ-2k20 (approved):
Good job 👍

PR #873, opened 2022-05-30
Develop

MabuzaM:
[DEMO LINK](https://MabuzaM.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,15 +1,34 @@
 import React from 'react';
 import './App.scss';
+import Sum from './components/Sum';
+import SumProps from './components/SumProps';
 
 const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum
+      a={SumProps.sum1.a}
+      b={SumProps.sum1.b}
+    />
+
+    <Sum
+      a={SumProps.sum2.a}
+      b={SumProps.sum2.b}
+    />
+
+    <Sum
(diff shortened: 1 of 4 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+export const Sum = (props) => {
+  const { a, b } = props;
+
+  return (
+    <p>{`Sum of ${a} and ${b} is ${(a) + (b)}`}</p>
```
andreisucman, line 7: I think you don't need to interpolate if you're using a `<p>` tag

src/App.jsx
```diff
@@ -1,15 +1,34 @@
...
+
+    <Sum
+      a={SumProps.sum4.a}
+      b={SumProps.sum4.b}
+    />
+
+    <Sum
+      a={SumProps.sum5.a}
+      b={SumProps.sum5.b}
+    />
```
andreisucman, line 31: I did it in a similar way but then I learned that you can simply pass the integers like this `a={2}` `b={3}` and it becomes like 10 times shorter. Just for your info.

andreisucman (approved):
Good job, but please consider my comments.

src/components/Sum.jsx
MabuzaM: Okay noted thanks

src/App.jsx
MabuzaM: Okay noted thanks. Will attend to it

src/components/SumProps.jsx
```diff
@@ -0,0 +1,21 @@
...
+  sum3: {
+    a: 10,
+  },
+
+  sum4: {
+    b: 5,
+  },
+
+  sum5: {},
+};
```
ihor-jpeg, line 21: i think it's a bit too much

ihor-jpeg (changes requested):

src/components/SumProps.jsx
MabuzaM: Have fixed

VinnieJ-2k20 (approved):
Good job 👍

PR #876, opened 2022-05-30
Solution

amitpatiljc:
[DEMO LINK](https://amitpatiljc.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -4,10 +4,20 @@ import { Sum } from './components';
 
 const App = () => (
   <>
-    <Sum a={2} b={3} />
-    <Sum a={5} b={-5} />
-    <Sum a={10} />
-    <Sum b={5} />
+    <Sum
+      a={2}
+      b={3}
+    />
+    <Sum
+      a={5}
+      b={-5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
     <Sum />
   </>
 );


src/components/Sum.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = (props) => {
+  const { a, b } = props;
+
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
andreisucman, line 8: I think you don't need to interpolate inside a `<p>` tag.

andreisucman (approved):
Good job. But I have one comment.

src/components/Sum.jsx
ihor-jpeg: agree, but one linter rule must be disabled. it's best to do that in eslint config file

src/App.jsx
```diff
@@ -1,15 +1,14 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={5} b={-5} />
+    <Sum a={10} />
+    <Sum b={5} />
+    <Sum />
```
ihor-jpeg, line 11: let's put props on individual lines

ihor-jpeg (changes requested):

VinnieJ-2k20 (approved):
Good job 👍

PR #880, opened 2022-05-31
solution

iLebo-stack:
https://iLebo-stack.github.io/react_sum/

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+const Sum = ({ a, b }) => {
+  const aVal = a || 0;
+  const bVal = b || 0;
+
+  return (
+    <p>
+      Sum of {aVal} and {bVal} is {aVal + bVal}
+    </p>
+  );
+};
+
+export default Sum;
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Sum component
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_sum/)
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://iLebo-stack.github.io/react_sum/)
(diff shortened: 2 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const Sum = ({ a, b }) => {
+  const aValue = a || 0;
+  const bValue = b || 0;
```
ihor-jpeg, line 5: it's alright, but in TS this approach will not work, because a and b are both required

src/components/Sum.spec.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { mount } from '@cypress/react';
-import { Sum } from './Sum';
+import Sum from './Sum';
```
ihor-jpeg, line 3: you can use named exports and you wont have modify test files

ihor-jpeg (changes requested):

src/components/Sum.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const Sum = ({ a = 0, b = 0 }) => (
```
meforoma, line 3: follow Ihor's advice regarding named exports
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```

src/components/Sum.jsx
```diff
@@ -0,0 +1,9 @@
...
+  <p>
+    Sum of {a} and {b} is {a + b}
+  </p>
+);
+
+export default Sum;
```
meforoma, line 9:
```suggestion
```

src/components/Sum.spec.jsx
meforoma, line 3: named exports are preferred
```suggestion
import { Sum } from './Sum';
```

meforoma (changes requested):

meforoma (approved):

PR #881, opened 2022-05-31
react_sum

Sardorbek1122:
[DEMO LINK](https://Sardorbek1122.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,15 +1,14 @@
 import React from 'react';
 import './App.scss';
+import { Sum } from './components';
 
 const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={5} b={5} />
+    <Sum a={10} />
+    <Sum b={5} />
+    <Sum />
   </>
 );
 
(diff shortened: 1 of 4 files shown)


src/App.jsx
```diff
@@ -1,15 +1,14 @@
...
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={5} b={5} />
```
rHashTag, line 8: `a` should be `-5`.
```suggestion
    <Sum a={-5} b={5} />
```

rHashTag (changes requested):
Good job with the code. There is one error, check my comment.

src/components/Sum.jsx
```diff
@@ -0,0 +1,24 @@
...
+  a: 0,
+  b: 0,
+};
+
+Sum.propTypes = {
+  a: PropTypes.number,
+  b: PropTypes.number,
+};
+
+export default Sum;
```
meforoma, line 24: ? why export default if in line 4: `export const Sum`

src/components/Sum.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const Sum = ({ a, b }) => {
+  const res = a + b;
+
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${res}`}
+    </p>
+  );
+};
```
meforoma, line 12: `res` variable is not really of a much use here - `a + b` is shorter than declaring a variable for it
```suggestion
export const Sum = ({ a, b }) => (
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
);
```

meforoma (changes requested):

meforoma (approved):

PR #884, opened 2022-06-01
test

WojciechCzarnowski:


src/App.jsx
```diff
@@ -1,15 +1,14 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum a={2} and b={5} />
+    <Sum a={-5} and b={0} />
+    <Sum a={10} and b={10} />
+    <Sum a={0} and b={5} />
+    <Sum a={0} and b={0} />
```
proznerka, line 11: I think we need to add also `<Sum />` at the end.

proznerka (approved):

src/App.jsx
ihor-jpeg: agree, you should take care of calling Sum without any props

src/components/Sum.spec.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { mount } from '@cypress/react';
-import { Sum } from './Sum';
```
ihor-jpeg, line 3: you can avoid this if you use named exports

ihor-jpeg (changes requested):

src/App.jsx
```diff
@@ -1,15 +1,15 @@
...
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove comment lines :) */}
+    <Sum a={2} b={5} />
+    <Sum a={-5} b={0} />
```
meforoma, line 8: no need to pass zero value since you're already handling it when providing default value for a prop in `Sum` component
```suggestion
    <Sum a={-5} />
```

src/App.jsx
```diff
@@ -1,15 +1,15 @@
...
+    <Sum a={10} b={10} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
+    <Sum />
```
meforoma, line 12: same here, no need to pass `0`
```suggestion
    <Sum b={5} />
    <Sum />
```

meforoma (approved):
pass
yet, pay attention to fresh comments

PR #886, opened 2022-06-02
done

elkhayate:


amitpatiljc (approved):
Good job 👍

src/App.jsx
```diff
@@ -1,15 +1,25 @@
...
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
+    <Sum />
```
meforoma, line 22: over-multi-lined
make each a one-liner - it's totally appropriate here

src/components/Sum.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Sum = ({
```
meforoma, line 4: you should be able to export right here
```suggestion
export const Sum = ({
```
then you don't seem to need export at last line

src/components/Sum.jsx
```diff
@@ -0,0 +1,28 @@
...
+  a,
+  b,
+}) => (
+  <p>
+    <span>Sum of </span>
+    {a}
+    <span> and </span>
+    {b}
+    <span> is </span>
+    {a + b}
```
meforoma, line 14: string literal is a perfect solution here

```suggestion
    `Sum of ${a} ...`
```

meforoma (changes requested):

src/components/Sum.jsx
```diff
@@ -0,0 +1,23 @@
...
+import PropTypes from 'prop-types';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
+  <p>
+    <span>{`Sum of ${a}`}</span>
+    <span>{` and ${b}`}</span>
+    <span>{` is ${a + b}`}</span>
```
meforoma, line 11: why do you even need 3 spans?
raw string literal would do just fine
```suggestion
    {`Sum of ${a} and ${b} is ${a + b}`}
```

meforoma (changes requested):

meforoma (approved):

PR #896, opened 2022-06-26
React sum

vsupruniuk:
[DEMO LINK](https://vsupruniuk.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,12 @@
+import React from 'react';
+
+export const Sum = ({ a, b }) => (
+  <p>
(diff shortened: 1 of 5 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+export const Sum = ({ a, b }) => (
```
mykola-koval, line 3: will it work like this?
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```

mykola-koval (approved):

PR #901, opened 2022-06-29
Solution

mira1zu:
- [DEMO LINK](https://mira1zu.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
```
mykola-koval, line 4: will it work like this?
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```

mykola-koval (approved):

PR #902, opened 2022-06-29
solved

mflkvsk:
[DEMO LINK](https://mflkvsk.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,15 +1,14 @@
 import React from 'react';
 import './App.scss';
+import Sum from './components/Sum';
 
 const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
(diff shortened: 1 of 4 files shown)


src/App.jsx
```diff
@@ -1,15 +1,14 @@
 import React from 'react';
 import './App.scss';
+import { Sum } from './components';
```
Olexandr6, line 3: To import in such a way, you should create `index.js` file and `export *` from related file

src/App.scss
```diff
@@ -1 +1,7 @@
-// styles go here
+p {
+  width: 150px;
+  border: 2px solid darkorange;
+  border-radius: 5px;
+  text-align: center;
+  color: rebeccapurple;
+}
```
Olexandr6, line 7: Don't style by tag, and move this style to Sum.css

src/components/Sum.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
```
Olexandr6, line 1: Create a separate folder for this component. For example `src/components/Sum/Sum.jsx`

Olexandr6 (changes requested):
Almost done👍

Olexandr6 (changes requested):
Can't see all files, looks like you didn't commit them

Olexandr6 (approved):
Good solution👍

PR #1008, opened 2022-08-01
solution

vitmixG:
[DEMO LINK](https://vitmixG.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,13 @@
-// export function Sum() {}
+import React from 'react';
+
+const Sum = (props) => {
+  const { a = 0, b = 0 } = props;
+
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+};
+
+export default Sum;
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,13 @@
-// export function Sum() {}
+import React from 'react';
+
+const Sum = (props) => {
```
VETALEK, line 3:
```suggestion
export const Sum = (props) => {
```

you need the named export to pass the tests 😉

src/components/Sum.jsx
```diff
@@ -1 +1,13 @@
...
+  const { a = 0, b = 0 } = props;
+
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+};
+
+export default Sum;
```
VETALEK, line 13:
```suggestion
```
here you don't need the default export

VETALEK (changes requested):
Like 👍 
just need to fix exports

VETALEK (approved):
awesome 🤩

PR #1015, opened 2022-08-01
Solution

Lentyaeev:
[DEMO LINK](https://lentyaeev.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,11 @@
-// export function Sum() {}
+import React from 'react';
+
+function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
+
+export default Sum;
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,11 @@
-// export function Sum() {}
+import React from 'react';
+
+function Sum({ a = 0, b = 0 }) {
```
VETALEK, line 3:
```suggestion
export function Sum({ a = 0, b = 0 }) {
```
so you can pass the tests 😉

src/components/Sum.jsx
```diff
@@ -1 +1,11 @@
...
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
+
+export default Sum;
```
VETALEK, line 11: you need to use named export

VETALEK (changes requested):
GJ 👍 
commented

VETALEK (approved):
🤩

PR #1042, opened 2022-08-29
solve the reactSum problem

tsybulko-natalya:
https://tsybulko-natalya.github.io/react_sum/


src/components/Sum.jsx
```diff
@@ -1 +1,17 @@
...
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
```
marina-tilniak, line 15: let's use interpolation

marina-tilniak (changes requested):

PR #1046, opened 2022-08-30
Solution

PaulMaistrenko:
[DEMO LINK](https://PaulMaistrenko.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,22 @@
-// export function Sum() {}
+import React from 'react';
+
+const Sum = ({
+  a,
+  b,
+}) => (
+  <p>
+    Sum of
+    {` ${a} `}
+    and
+    {` ${b} `}
+    is
+    {` ${a + b}`}
+  </p>
+);
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
+
+export { Sum };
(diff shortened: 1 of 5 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,22 @@
...
+  a,
+  b,
+}) => (
+  <p>
+    Sum of
+    {` ${a} `}
+    and
+    {` ${b} `}
+    is
+    {` ${a + b}`}
```
marina-tilniak, line 13: let's use interpolation here

marina-tilniak (changes requested):

src/components/Sum.jsx
PaulMaistrenko: I fixed it.

marina-tilniak (changes requested):
If you have interpolation, you can use only one string

PaulMaistrenko:
> If you have interpolation, you can use only one string

image
if i use only one string i have some mistake frome linter. I can disable this error.

src/components/Sum.jsx
```diff
@@ -1 +1,18 @@
-// export function Sum() {}
+/* eslint-disable react/jsx-one-expression-per-line */
+import React from 'react';
+
+const Sum = ({
+  a,
+  b,
+}) => (
+  <p>
+    Sum of{` ${a} `}and{` ${b} `}is{` ${a + b} `}
```
Bohdan-Maksymiuk, line 9:
```suggestion
    {`Sum of ${a} and ${b} is ${a + b}`}
```
you may ask in fe_chat if smth remains unclear

Bohdan-Maksymiuk (changes requested):

src/components/Sum.jsx
PaulMaistrenko: corrected

marina-tilniak (approved):

PR #1049, opened 2022-09-01
Solution

ostUp:
[DEMO LINK](https://ostUp.github.io/react_sum/)


Roman-Artiukh:
image
last line is extra))

src/components/Sum.jsx
```diff
@@ -1 +1,13 @@
-// export function Sum() {}
+import React from 'react';
+
+function Sum({ a = 0, b = 0 }) {
+  return (
+    <>
+      <p>
+        {`Sum of ${a} and ${b} is ${a + b}`}
+      </p>
+    </>
```
AndriiRut, line 9:
```suggestion
      <p>
        {`Sum of ${a} and ${b} is ${a + b}`}
      </p>
```
don't need of React.Component here.
whatever you do, your component must return only one parent element(div, span, p, `<>` and etc.). No mater how many elements are in it, The parent element should be only the one! and here you already have one the `<p>`

AndriiRut (changes requested):

AndriiRut (approved):
🚀

PR #1068, opened 2022-09-05
Sum solution

ibondrk:


NataliyaAn (approved):
completely great job!

src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
...
+
+export function Sum({ a, b }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
+
+Sum.defaultProps = {
```
AndriiRut, line 11: good! also, you can add default value as you did it in functions

`export function Sum({ a = 0, b = 0}) {...}`

AndriiRut (approved):
🚀

src/components/Sum.jsx
ibondrk: Thanks, i'll keep it in mind. There was a lot of variations, and i decide to do it this way, but your rigth, it will be more compact

PR #1129, opened 2022-09-28
Develop

Quelaan1:
- [DEMO LINK](https://Quelaan1.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,19 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,19 @@
...
+    />
+    <Sum
+      a={-5}
+      b={5}
+    />
+    <Sum a={10} />
+    <Sum b={5} />
+    <Sum />
   </>
 );
```
kamal-tufekcic, line 19: Just a note, unlike with raw HTML, in React linter allows up to 3 attributes per line, so no need for newlines here

kamal-tufekcic (approved):

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #1159, opened 2022-10-11
Wrote a function sum in component sum

Yeliseiev:


Maksym845 (approved):
Well done👍

src/components/Sum.jsx
```diff
@@ -1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
```
Denchepornyuk, line 3: add default values to the defaultProps object

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1166, opened 2022-10-11
React sum solution

Yevheniya-Deryka:
[DEMO LINK](https://Yevheniya-Deryka.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+);
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+);
```
Anna-Borodaienko, line 5:
```suggestion
export const Sum = ({ a, b }) => (
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
   </p>
);
```
I think this sign would be more readable

Anna-Borodaienko (approved):

src/components/Sum.jsx
Yevheniya-Deryka: agree, thanks! fixed

naziamoff (approved):
🔥

PR #1187, opened 2022-10-12
React sum

glugovsky:
[DEMO LINK](https://glugovsky.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${Number(a) + Number(b)}`}</p>
+  );
+}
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${Number(a) + Number(b)}`}</p>
```
OleksandrGlulhovskiy, line 5: Do you need Number(a) and Number(b), you can use just a and b, but i think it's not a mistake

OleksandrGlulhovskiy (commented):
Perfect!!!

src/components/Sum.jsx
naziamoff: Oleksandr is right, those variables will be numbers anyway.
What is more - it's better to make components an arrow function

naziamoff (changes requested):

src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
naziamoff, line 4:
```suggestion
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
```

naziamoff (approved):
🔥

PR #1191, opened 2022-10-12
react_sum

Arsen-Shpak:
DEMO LINK

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,8 @@
 // export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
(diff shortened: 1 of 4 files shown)


TheVovchik (approved):
Demo not clickable in description

code is good

src/components/Sum.jsx
```diff
@@ -1 +1,8 @@
 // export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
Denchepornyuk, line 4: add default values to the defaultProps object

Denchepornyuk (changes requested):
fix your demo link

OleksiiBlezniuk (approved):

PR #1193, opened 2022-10-12
Done

iPanda177:
[DEMO LINK](https://iPanda177.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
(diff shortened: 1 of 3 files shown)


iam-datsenko (approved):

src/components/Sum.jsx
```diff
@@ -1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
```
naziamoff, line 4: you can do return statement right away if you don't need any actions before the return statement

naziamoff (changes requested):

src/components/Sum.jsx
naziamoff: and one more thing - we use arrow functions in functional components

naziamoff (approved):
🔥

PR #1233, opened 2022-11-03
React sum task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,13 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({
+  a = 0,
+  b = 0,
+  text = `Sum of ${a} and ${b} is ${a + b}`,
+}) {
+  return (
+    <p>
+      {text}
+    </p>
+  );
+}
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,13 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({
+  a = 0,
+  b = 0,
+  text = `Sum of ${a} and ${b} is ${a + b}`,
```
Bohdan-Maksymiuk, line 6: you didn't accept this string in props - so move it inside component as a variable or use in return directly

Bohdan-Maksymiuk (changes requested):

src/components/Sum.jsx
juliiaap: fixed

src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({
+  a = 0,
+  b = 0,
+}) {
```
Anastasiia-Svintsova, line 6:
```suggestion
export function Sum({ a = 0, b = 0 }) {
```

Anastasiia-Svintsova (approved):

PR #1239, opened 2022-11-12
Solution

bokorzita:
https://bokorzita.github.io/react_sum/

### DIFF
--- a/src/components/Sum/Sum.jsx
+++ b/src/components/Sum/Sum.jsx
@@ -1,7 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export const Sum = props => <p>{`Sum of ${props.a} and ${props.b} is ${props.a + props.b}`}</p>;
+export const Sum = (props) => {
+  const { a, b } = props;
+
+  return <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
+};
 
 Sum.defaultProps = {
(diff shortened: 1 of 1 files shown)


src/components/Sum/Sum.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = props => <p>{`Sum of ${props.a} and ${props.b} is ${props.a + props.b}`}</p>;
```
amitpatiljc, line 4: we can some code by destructring 
`const { a, b } = props`

amitpatiljc (changes requested):
Good Job! Just one small change needed

src/components/Sum/Sum.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = (props) => {
+  const { a, b } = props;
+
+  return <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
+};
```
qoa11a, line 8:
```suggestion
export const Sum = ({ a, b }) => {
  return <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
};
```
You can do it right here. Just a hint. :)

qoa11a (approved):

PR #1244, opened 2022-11-14
added Sum components

Artem66:
[DEMO LINK](https://Artem66.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
+  (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  )
+);
+
+export { Sum };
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+const Sum = ({
```
iryna-lelyk, line 3: When you use named export, you can export it right before your function
```suggestion
export const Sum = ({
```

src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
...
+  a = 0,
+  b = 0,
+}) => (
+  (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  )
+);
```
iryna-lelyk, line 10: You don't need additional round brackets if you don't write word `return ` 
```suggestion
}) => (
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
);
```

PR #1259, opened 2022-11-15
solution

andrii-bassarab:
[DEMO LINK](https://andrii-bassarab.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,13 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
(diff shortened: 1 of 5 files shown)


src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
sofiia-dovhan, line 11: According to the task 
Where attribute value is 0 it means that there has to be no argument and in this case has to work your default value = 0
remove it from App.jsx

sofiia-dovhan (changes requested):
Gj, but you have to check your code with task

sofiia-dovhan (approved):
:fire:

mykola-koval (approved):

PR #1275, opened 2022-11-16
add solution

AlexTsimba:
[DEMO LINK](https://alextsimba.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,20 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum(props) {
+  return (
+    <p clasName="paragraph">
+      {`Sum of ${props.a} and ${props.b} is ${props.a + props.b}`}
+    </p>
+  );
+}
+
+Sum.defaultProps = {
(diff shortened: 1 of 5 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,20 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum(props) {
+  return (
+    <p clasName="paragraph">
```
Oleksandr-Stryzhavchuk, line 6: clasName is not needed here

Oleksandr-Stryzhavchuk (approved):
Good job, only a few corrections

src/components/Sum.jsx
```diff
@@ -1 +1,20 @@
...
+      {`Sum of ${props.a} and ${props.b} is ${props.a + props.b}`}
```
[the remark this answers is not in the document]
illiarerun: agree

src/components/Sum.jsx
illiarerun: agree

illiarerun (changes requested):

PR #1280, opened 2022-11-16
solution

MelnychukYana:
https://MelnychukYana.github.io/react_sum/


src/components/Sum.jsx
```diff
@@ -1 +1,18 @@
...
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  </>
+);
+
+Sum.propTypes = {
+  a: PropTypes.number.isRequired,
+  b: PropTypes.number.isRequired,
+};
```
SanyaBratashchuk, line 18:
```suggestion
Sum.defaultProps {
  a: 0,
  b: 0,
}

Sum.propTypes = {
  a: PropTypes.number,
  b: PropTypes.number,
};
```

src/components/Sum.jsx
```diff
@@ -1 +1,18 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
```
SanyaBratashchuk, line 6:
```suggestion
  a,
  b,
```
unnecessary if you set defaultProps

SanyaBratashchuk (changes requested):

PR #1303, opened 2022-12-07
Implemented the solution

zynkro:
[DEMO LINK](https://zynkd.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
 export function Sum({ a, b }) {
   const sum = a + b;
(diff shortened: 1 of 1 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum(props) {
+  const { a = 0, b = 0 } = props;
```
polosanya, line 4:
```suggestion
export function Sum({ a = 0, b = 0 }) {
```
You can immediately get them using [destructuring](https://www.geeksforgeeks.org/destructuring-of-props-in-reactjs/)

polosanya (approved):
Good job! Please pay attention to my comment, it might be useful)

PR #1305, opened 2022-12-11
Solution

shyptia:
[DEMO LINK](https://shyptia.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,8 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,8 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
Oleksandr-Filo, line 7:
```suggestion
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
```

Oleksandr-Filo (approved):
looks good, but for me it is better to use multiline format of HTML element

solaryasha (approved):
Good job!

PR #1332, opened 2022-12-22
I could change the fate of people, but I just change the variables...

RedkoSerhii:
[DEMO LINK](https://RedkoSerhii.github.io/react_sum/) and add it to the PR description

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
+  <p>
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
+  <p>
```
Sergii-Osedchuk, line 7: According to check list we have to write arguments in one line if there are less than three. But anyway good job!!!

Sergii-Osedchuk (approved):
Make changes with arguments. Everything else is ok!!!

solaryasha (approved):
👍

PR #1335, opened 2022-12-22
Solution

skorykdariia:
[DEMO LINK](https://skorykdariia.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
Mykhailyk-Vadym, line 6: I guess we should write in an 1 row as it sad in Checklist p1

solaryasha (approved):
👍

PR #1348, opened 2022-12-22
Added Sum Component

viktoriia-kravchenko:
[DEMO LINK](https://viktoriia-kravchenko.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    {`Sum of ${a} and ${b} is ${a + b}`}
+  </p>
+);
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
Sveta-Kryukova, line 3: Sum.defaultProps = {
  a: 0,
  b: 0,
}

I think we shoul use this for default in this task

Sveta-Kryukova (changes requested):

src/components/Sum.jsx
viktoriia-kravchenko: I believe we can destructure the props of Sum with default values without defaultProps since our default values are short enough as well as the properties themselves.

Sveta-Kryukova (approved):

src/components/Sum.jsx
valerii-derkach: let's add `default props` and `prop types `to learn how they work

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1351, opened 2022-12-22
sum

kreisberg:


src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
Sveta-Kryukova, line 11: I think we shouldn't write zero if we set a default property

src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({
+  a = 0,
+  b = 0,
+}) {
```
Sveta-Kryukova, line 6: also I think we should use this for default arguments 
Sum.defaultProps = {
  a: 0,
  b: 0,
}

Sveta-Kryukova (changes requested):

src/App.jsx
kreisberg: ✔️

src/components/Sum.jsx
kreisberg: Fixes were added. Thanks!

src/components/Sum.jsx
```diff
@@ -1 +1,23 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum({
+  a = 0,
+  b = 0,
+}) {
```
valerii-derkach, line 7:
```suggestion
export function Sum({ a, b }) {
```

if you have default props, you do not need to assign them here

valerii-derkach (changes requested):

src/components/Sum.jsx
kreisberg: ✔️

valerii-derkach (approved):

PR #1359, opened 2022-12-23
React Sum

Ruslan-Pashkov:
[DEMO LINK](https://ruslan-pashkov.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,10 +1,18 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-export const Sum = ({
-  a = 0,
-  b = 0,
-}) => (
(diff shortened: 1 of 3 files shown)


Sveta-Kryukova (changes requested):
Good, but add Sum.propTypes for a and b=)

src/components/Sum.jsx
```diff
@@ -1 +1,18 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
```
Ihor-Karpyn, line 4: better to make the destructuring inside the body of the function

Ihor-Karpyn (approved):
GJ

PR #1363, opened 2022-12-25
Add solution

VladimirEdelmann:
# Task
Create a `Sum` component accepting `a` and `b` props and rendering a paragraph with a text like `Sum of 2 and 1 is 3`. Replace numbers with actual
values. If the `a` or `b` props are not passed, set them to `0`.

**Please note:** use named export (not default).

The `App` should contain the next `Sum` components:

- `a = 2` and `b = 3`;
- `a = -5` and `b = 5`;
- just `a = 10`;
- just `b = 5`;
- no params at all.

# Implementation
[DEMO LINK](https://vladimiredelmann.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,14 @@
 import React from 'react';
+import { Sum } from './components/Sum';
+
 import './App.css';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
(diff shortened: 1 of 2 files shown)


DimaKolesnyk (approved):
To the moon  🚀

valerii-derkach (changes requested):
good job!

let's add default props and propTypes

valerii-derkach (approved):

PR #1365, opened 2022-12-25
Solution

novytskyi-d:
[DEMO LINK](https://novytskyi-d.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
+  <p>
+    {`Sum of ${a} and ${b} is ${a + b}`}
+  </p>
+);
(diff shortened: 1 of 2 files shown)


viktoriia-kravchenko (approved):
Good job!

src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
Ihor-Karpyn, line 3: Better to make destructuring inside body of the function

src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
...
+  <p>
+    {`Sum of ${a} and ${b} is ${a + b}`}
+  </p>
+);
```
Ihor-Karpyn, line 7:
```suggestion
export const Sum = (props) => {
const { a = 0, b = 0 } = props;

return (
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
)};
```

Ihor-Karpyn (approved):
GJ

PR #1368, opened 2022-12-26
Solution

TetKopot:


src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
kreisberg, line 3: It's better way to check new value ('a' and 'b') on new string.

kreisberg (approved):

valerii-derkach (approved):

PR #1373, opened 2022-12-26
Solution

Rijo-J:
- [DEMO LINK](https://Rijo-J.github.io/react_person/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,14 @@
 // export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+);
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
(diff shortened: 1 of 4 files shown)


nurdanokcu (approved):
:white_check_mark:

naziamoff (changes requested):
the demo is from different task

Rijo-J:
> the demo is from different task
✅ 
My bad 😬 sorry!

src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
 // export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
naziamoff, line 8: It's better to move the text to the new line, it is more readable this way
```suggestion
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
```

src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
...
+);
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
```
naziamoff, line 14: why not just assign it inside the arguments?

naziamoff (approved):
LGTM! Just a little bit of polishing would be nice

PR #1375, opened 2022-12-26
solution

Cardlice:
[DEMO LINK](https://Cardlice.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => (
```
Ihor-Karpyn, line 3: Better to make the destructuring inside bode of the function

Ihor-Karpyn (approved):
GJ

PR #1383, opened 2022-12-27
finished

jackelashkar:
https://jackelashkar.github.io/react_sum/

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a, b }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  );
+}
+
+Sum.defaultProps = {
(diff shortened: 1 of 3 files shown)


alaaashkar (approved):
nice !

src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a, b }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  );
+}
```
naziamoff, line 7: - better to make functional components an arrow functions
- you should return right away if you don't have any calculations inside your arrow function
- better to move text to a separate line, it improves readability
```suggestion
export const Sum = ({ a, b }) => (
    <p>
      {`Sum of ${a} and ${b} is ${a + b}`}
    </p>
);
```

naziamoff (approved):
looks good, but there are some suggestions

PR #1384, opened 2022-12-27
Develop

alaaashkar:
[DEMO LINK](https://alaaashkar.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  );
+}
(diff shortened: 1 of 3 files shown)


jackelashkar (approved):
awesome!

src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  );
+}
```
naziamoff, line 7: - make it an arrow function
- return right away (use `()` instead of `{}`)
- let's move the text to the new line

naziamoff (changes requested):

src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
naziamoff, line 4: would be better to move the text to the new line

naziamoff (approved):

PR #1388, opened 2022-12-27
Finished

JzSol:
[DEMO LINK](https://JzSol.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,8 +1,10 @@
 import React from 'react';
 
-export const Sum = ({ a, b }) => <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
-
-Sum.defaultProps = {
-  a: 0,
-  b: 0,
-};
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
+  <p>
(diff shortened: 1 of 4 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,8 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
```
dolnaro, line 8: First; you should move the <p> to the new line.

Second; I don't think you need a defaultProp here because you can give default values on the upper lines.
When you are adding { a, b } params, you can give them default values. And by doing that, you should also move them to the following lines.

dolnaro (changes requested):

src/components/Sum.jsx
JzSol: You are right, Thank you!😄

dolnaro: Happy to help :)

naziamoff (approved):
🔥

PR #1396, opened 2023-01-03
Solution by DK

DimaK123:
[DEMO LINK](https://DimaK123.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,22 +1,9 @@
 import React from 'react';
 
-export function Sum({
-  a = 0,
-  b = 0,
-}) {
+export function Sum({ a = 0, b = 0 }) {
   return (
     <p>
-      Sum of
-      {' '}
-      {a}
-      {' '}
-      and
-      {' '}
-      {b}
-      {' '}
-      is
-      {' '}
-      {a + b}
+      {`Sum of ${a} and ${b} is ${a + b}`}
     </p>
   );
 }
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,22 @@
...
+      {a}
+      {' '}
+      and
+      {' '}
+      {b}
+      {' '}
+      is
+      {' '}
+      {a + b}
+    </p>
```
polosanya, line 20: image

src/components/Sum.jsx
```diff
@@ -1 +1,22 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({
+  a = 0,
+  b = 0,
+}) {
```
polosanya, line 6:
```suggestion
export function Sum({ a = 0, b = 0 }) {
```
You can leave it in one line as there are < 3 props and the line is quite short

polosanya (changes requested):
Almost done! Please check the comments

src/components/Sum.jsx
DimaK123: Fixed

src/components/Sum.jsx
DimaK123: Fixed

src/components/Sum.jsx
```diff
@@ -1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
```
OlEzhA0, line 9: for future: use arrow functions to make your code prettier and shorter 
```suggestion
export const Sum = ({ a = 0, b = 0 })  => (
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
);
```

OlEzhA0 (approved):
GJ!

PR #1405, opened 2023-01-11
solution

illia-kyselov:


src/App.jsx
```diff
@@ -1,14 +1,22 @@
...
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum
+      a={2}
+      b={3}
+    />
```
TarasHoliuk, line 10: if there less than 3 args, and the line length is < 80 (the number depends on the linter setting, 80 is the current value) you can write it in this way (FYI):

```suggestion
    <Sum a={2} b={3} />
```

TarasHoliuk (approved):

PR #1410, opened 2023-01-16
initial

kostasburnazaki:
[DEMO LINK](https://kostasburnazaki.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,20 +1,14 @@
 import React from 'react';
 
-export const Sum = ({ a, b }) => (
-  <p>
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,22 @@
...
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
+  </p>
```
marina-tilniak, line 16: let's use interpolation here

marina-tilniak (changes requested):

PR #1415, opened 2023-01-18
task done

Zykhor-Andrii:
[DEMO LINK](https://Zykhor-Andrii.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,17 @@
-// export function Sum() {}
+
+import React from 'react';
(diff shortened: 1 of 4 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,17 @@
-// export function Sum() {}
+
+import React from 'react';
```
tetiana-ivanchenko, line 2:
```suggestion
import React from 'react';
```

tetiana-ivanchenko (approved):

PR #1422, opened 2023-01-20
Paragraph content was rewrite

anna-inozemets:
[DEMO LINK](https://anna-inozemets.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,14 @@
 import React from 'react';
+import { Sum } from './components/Sum';
+
 import './App.css';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a="2" b="3" />
+    <Sum a="-5" b="5" />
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -1,14 +1,14 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a="2" b="3" />
+    <Sum a="-5" b="5" />
+    <Sum a="10" b="0" />
+    <Sum a="0" b="5" />
+    <Sum a="0" b="0" />
```
Avichez, line 12: You have set up default values in Sum component, so it's not necessary to pass props with `0`.

Avichez (approved):
Great job 🙂 👍 
Just a small comment.

src/App.jsx
Anna-Tkachenko: agree, fix @Avichez comment

src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
+import React from 'react';
 // export function Sum() {}
+export const Sum = ({ a, b }) => (
+  <p>
+    {`Sum of ${+a} and ${+b} is ${+a + +b}`}
```
Anna-Tkachenko, line 5: use prop types property to specify correct types for component arguments instead of using `+` to transforming them to number

Anna-Tkachenko (changes requested):

alexandra-protyanova (approved):
🚀

PR #1434, opened 2023-01-22
finished react-sum task with passed tests

andrii-kurash:
All the best!🔥

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  );
+}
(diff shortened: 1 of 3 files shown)


NVyrsta (approved):
Great :)

src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  );
+}
```
Denchepornyuk, line 7: use PropTypes for Sum component
Also, use arrow function for components

Denchepornyuk (changes requested):

src/components/Sum.jsx
andrii-kurash: Okay👌

Denchepornyuk (approved):

PR #1462, opened 2023-01-23
completed

AlexPro100pro:
[DEMO LINK](https://AlexPro100pro.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+);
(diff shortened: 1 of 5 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
anna-inozemets, line 3: Maybe it is better touse deafult props here
And also, don't forget to add prop  types

anna-inozemets (approved):
Really good code, just small comments!)

src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
...
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+);
```
Denchepornyuk, line 5: use PropTypes for this component, and default props also needed here

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1464, opened 2023-01-24
solution added

artem-sobolevskyi:
[DEMO LINK](https://artem-sobolevskyi.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
anna-inozemets, line 3: Maybe it is better to use deafultProps here

anna-inozemets (approved):
Pretty good!)

alexandra-protyanova (changes requested):
Add prop-types

alexandra-protyanova (approved):
GJ!

PR #1471, opened 2023-01-25
Solution

SashaViun:
[DEMO LINK](https://SashaViun.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,17 @@
 // export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,17 @@
 // export function Sum() {}
```
NVyrsta, line 1: don't forget to delete redundant comments

NVyrsta (approved):
Good job :0)

src/components/Sum.jsx
alexandra-protyanova: delete it

alexandra-protyanova (approved):
Well done 👍

PR #1472, opened 2023-01-25
sol 1.0

aliliolek:
[DEMOLINK](https://aliliolek.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => (
+  <>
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => (
+  <>
```
agniya-i, line 4: in this case not necessary to wrap into fragment `<>` 
```suggestion
```

agniya-i (changes requested):

Anastasiia-Svintsova (approved):

PR #1475, opened 2023-01-27
Initial commit

yaroslav-furhala:
- [DEMO LINK](https://yaroslav-furhala.github.io/react_sum/)


loralevitska (changes requested):
Ooops. your link is broken)
Please redeploy your solution, and before review check demo_link. If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

PR #1485, opened 2023-02-03
Finish task

U-Ra-All:
https://U-Ra-All.github.io/react_sum/

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    Sum of {a} and {b} is {a + b}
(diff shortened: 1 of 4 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    Sum of {a} and {b} is {a + b}
+  </p>
+);
```
mykhalenych, line 7: add proptypes

mykhalenych (changes requested):

src/components/Sum.jsx
U-Ra-All: Done

nazarmatsevych (approved):

PR #1505, opened 2023-02-16
commit message

Peti51:


Peti51:
[DEMO LINK](https://Peti51.github.io/react_sum/)

src/App.jsx
```diff
@@ -1,14 +1,21 @@
...
+      {' '}
+      and
+      {' '}
+      {b}
+      {' '}
+      is
+      {' '}
+      {a + b}
+      {' '}
+    </p>
```
qoa11a, line 19: - you need to move the p tag inside the Sum component
- you need to calculate a + b and show text inside the Sum component as well

src/index.jsx
```diff
@@ -2,7 +2,64 @@ import React from 'react';
...
+
+export const numbers4 = {
+  a: 0,
+  b: 5,
+};
+
+export const numbers5 = {
+  a: 0,
+  b: 0,
+};
```
qoa11a, line 28: - you shouldn't declare objects with numbers. Just pass them right away as arguments

src/index.jsx
```diff
@@ -2,7 +2,64 @@ import React from 'react';
...
+
+numbers4.defaultProps = {
+  a: 0,
+  b: 0,
+};
+
+numbers5.defaultProps = {
+  a: 0,
+  b: 0,
+};
```
qoa11a, line 65: - no need for this if you don't create these objects, as I mentioned before

src/index.jsx
```diff
@@ -2,7 +2,64 @@ import React from 'react';
...
+
 ReactDOM.render(
-  <App />,
+  <div>
+    <App {...numbers1} />
+    <App {...numbers2} />
+    <App {...numbers3} />
+    <App {...numbers4} />
+    <App {...numbers5} />
+  </div>,
```
qoa11a, line 37: - use React.Fragment instead of div here
- you should write your code inside App.jsx and use Sum.jsx inside of it

qoa11a (changes requested):
pay attention to the comments. Also, try to pass all the tests

Peti51:
Somehow npm test and npm start didnt work so this is the closest I could get.

src/components/Sum.jsx
```diff
@@ -1 +1,24 @@
-// export function Sum() {}
+import React from 'react';
+
+function Sum({ a, b }) {
```
polosanya, line 3: I think it's better to use arrow function, just like in App component.
You can also add default values for a and b

polosanya (changes requested):
Almost done! Please fix all the problems
image
Feel free to ask for help in chat if you have any difficulties)

src/components/Sum.jsx
```diff
@@ -1 +1,13 @@
...
+import React from 'react';
+
+const Sum = ({ a = 0, b = 0 }) => {
+  const result = a + b;
+
+    <>
+      <p>
+        {`Sum of ${a} and ${b} is ${result}`}
+      </p>
+    </>;
```
Volodymyr-Melnyk-Fron10der, line 10: 1. remove frament
2. I think you have to return this jsx from function

Volodymyr-Melnyk-Fron10der (changes requested):
image
Nothing wasn't rendered

qoa11a (approved):

PR #1515, opened 2023-02-22
solution

anna-tovkailo:


src/components/Sum.jsx
```diff
@@ -1 +1,17 @@
...
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
```
OleziO, line 15: Use template strings for concatenation strings and other data types.

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

OleziO (changes requested):

qoa11a (approved):

PR #1534, opened 2023-03-03
Initial solution

Dynamizx:
https://dynamizx.github.io/react_sum/

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -2,6 +2,6 @@ import React from 'react';
 
 export const Sum = ({ a = 0, b = 0 }) => (
   <p>
-    {`Sum of ${a} and ${b} is ${Number(a) + Number(b)}`}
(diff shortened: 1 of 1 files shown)


Volodymyr-Melnyk-Fron10der (changes requested):
Good job👍 But provide DEMO LINK in the PR description + make sure it's really working well
image

polosanya (approved):
Well done!

src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    {`Sum of ${a} and ${b} is ${Number(a) + Number(b)}`}
```
polosanya, line 5:
```suggestion
    {`Sum of ${a} and ${b} is ${a + b}`}
```
I think there is no need to convert a and b into numbers, as you pass only numbers

PR #1567, opened 2023-03-10
creat component Sum

mstrilec:
[DEMO LINK](https://mstrilec.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
```
DenysDubas, line 3: good point, but also there
```
Sum.defaultProps = {
  a: 0,
  b: 0,
};
```
if we have many default values

DenysDubas (approved):

Andr1yK (approved):

PR #1569, opened 2023-03-10
solution

Lilu580:
[DEMO LINK](https://Lilu580.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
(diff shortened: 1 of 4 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
stanislavkulakovskyi, line 4: not critical, but maybe it's better to format it like this: 
```suggestion
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
```

stanislavkulakovskyi (approved):
Good job!

Andr1yK (approved):

PR #1575, opened 2023-03-10
added solution

Vlados3611:


src/App.jsx
```diff
@@ -1,13 +1,14 @@
...
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
MaksymMaliuk, line 11: image

src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import { React } from 'react';
```
MaksymMaliuk, line 1: You don't need to use braces here.

src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
...
+export const Sum = ({ a, b }) => (
+  <p>
+    {`Sum of ${a} and ${b} is ${a + b}`}
+  </p>
+);
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
```
MaksymMaliuk, line 12: it works but it would be better if you set default parameters into destructurisation 
```
export const Sum = ({ a = 0, b = 0 })
```

MaksymMaliuk, line 12:
```suggestion
Sum.defaultProps = {
  a: 0,
  b: 0,
};
```
it works but it would be better if you set default parameters into destructurisation 
```
export const Sum = ({ a = 0, b = 0 })
```

MaksymMaliuk, line 12:
```suggestion
export const Sum = ({ a = 0, b =0 })
```

MaksymMaliuk (approved):

Andr1yK (approved):

PR #1579, opened 2023-03-11
added src and readme

DenysDubas:


src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum(args) {
```
Andr1yK, line 3: in react we usually name it `props`

Andr1yK (approved):

src/components/Sum.jsx
DenysDubas: okey

PR #1580, opened 2023-03-11
solution

anna-akhalmosulishvili:
https://anna-akhalmosulishvili.github.io/react_sum/

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,22 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum(props = 0) {
+  const aNum = +props.a || 0;
+  const bNum = +props.b || 0;
+
+  return (
+    <p>
+      Sum of
+      {' '}
+      {aNum}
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,22 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum(props = 0) {
+  const aNum = +props.a || 0;
+  const bNum = +props.b || 0;
```
Volodymyr-Melnyk-Fron10der, line 5: props object is always there
```suggestion
export function Sum({ a = 0, b = 0 }) {
  const aNum = +props.a;
  const bNum = +props.b;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Good, just one fix I would ask about

Volodymyr-Melnyk-Fron10der (approved):
Cool

VitaliyBodnar (approved):

PR #1597, opened 2023-03-15
Solution

Dmytro-Litinskyi:
[DEMO LINK](https://Dmytro-Litinskyi.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,11 @@
 // export function Sum() {}
+import React from 'react';
+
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,11 @@
 // export function Sum() {}
```
tetiana-ivanchenko, line 1:
```suggestion
```

tetiana-ivanchenko (approved):

PR #1600, opened 2023-03-15
React sum

ruslan-yarosh:
[DEMO LINK](https://ruslan-yarosh.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,27 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,27 @@
...
+
+    <Sum
+      a={10}
+    />
+
+    <Sum
+      b={5}
+    />
+
+    <Sum />
```
loralevitska, line 25: If the length of the code is less than 80-100 characters (depending on the linter settings), it is better to write in a line so as not to inflate the code
```suggestion
    <Sum a={2} b={3}  />
    <Sum a={-5} b={5} />
    <Sum a={10}  />
    <Sum b={5} />
    <Sum />
```

src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({
+  a = 0,
+  b = 0,
+}) {
```
loralevitska, line 6: Same here
```suggestion
export function Sum({ a = 0, b = 0 }) {
```

loralevitska (approved):

PR #1607, opened 2023-03-23
Dev

NazikVr:
[DEMO LINK](https://nazikvr.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -4,31 +4,14 @@ import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
+    <Sum a={2} b={3} />
 
-    <Sum
-      a={2}
-      b={3}
-    />
+    <Sum a={-5} b={5} />
 
-    <Sum
-      a={-5}
-      b={5}
-    />
+    <Sum a={10} b={0} />
 
-    <Sum
-      a={10}
-      b={0}
-    />
+    <Sum a={0} b={5} />
 
-    <Sum
-      a={0}
-      b={5}
-    />
-
-    <Sum
-      a={0}
-      b={5}
-    />
-
+    <Sum a={0} b={5} />
   </>
 );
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,34 @@
...
+
+    <Sum
+      a={0}
+      b={5}
+    />
+
+    <Sum
+      a={0}
+      b={5}
+    />
```
polosanya, line 31:
```suggestion
    <Sum a={0} b={0} />
```
In last one should be no params at all
image
image

src/App.jsx
```diff
@@ -1,14 +1,34 @@
...
+
   </>
```
polosanya, line 33:
```suggestion
    />
  </>
```
Don't leave empty lines between parent and child elements

src/components/Sum.jsx
```diff
@@ -1 +1,17 @@
...
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
+  </p>
```
polosanya, line 16: image

polosanya (changes requested):
Almost done!

src/App.jsx
```diff
@@ -1,14 +1,17 @@
...
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+
+    <Sum a={-5} b={5} />
+
+    <Sum a={10} b={0} />
+
+    <Sum a={0} b={5} />
+
+    <Sum a={0} b={5} />
```
Volodymyr-Melnyk-Fron10der, line 15:
```suggestion
    <Sum a={10} />

    <Sum b={5} />

    <Sum />
```

Volodymyr-Melnyk-Fron10der (changes requested):
You shouldn't pass prop if it equals to zero

polosanya (changes requested):
Pay attention to task description. Last sum should be without parameters at all. 
Now it is just a copy of previous one
image
image

lerastarynets (approved):
All requirements are met. Keep up the great work!

PR #1612, opened 2023-03-28
react_sum done

EhorSaprykin:
- [DEMO LINK] (https://ehorsaprykin.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,20 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
+  <p>
+    Sum of
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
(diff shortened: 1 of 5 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,20 @@
...
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
+  </p>
```
tetiana-ivanchenko, line 19: use template literal

tetiana-ivanchenko (changes requested):

src/components/Sum.jsx
```diff
@@ -1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
polosanya, line 6:
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```
You can leave it in one line as there are < 3 props and the line is quite short

polosanya (approved):
Well done!

PR #1618, opened 2023-03-30
react sum sollution

serbinm:
[DEMO LINK](https://serbinm.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,13 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <p><Sum a="2" b="3" /></p>
```
agniya-i, line 7:
```suggestion
    <p><Sum a={2} b={3} /></p>
```

you can pass numbers in this way to not convert it later

src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+export function Sum({ a = 0, b = 0 }) {
+  const sum = +a + +b;
+
+  return `Sum of ${a} and ${b} is ${sum}`;
```
agniya-i, line 4:
```suggestion
  return <p>`Sum of ${a} and ${b} is ${sum}`</p>;
```

better move the tag inside the component, so you can use it without wrapping each time into `<p>`

agniya-i (changes requested):
Please fix demo link

PR #1623, opened 2023-03-31
complete task

VladGraund:
[DEMO LINK](https://VladGraund.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,17 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    Sum of
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
+  </p>
+);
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,17 @@
...
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
+  </p>
```
IvanFesenko, line 16:
```suggestion
  <p>Sum of {a} and {b} is {a + b}</p>
```

IvanFesenko (changes requested):

src/components/Sum.jsx
IvanFesenko: Make it in one line

VladGraund: Issue

linter does not allow me to do so

VladGraund: and when I commit a fix code, linter brings my code back

anstsot, line 16: to avoid the linter errors you can use template literals
```suggestion
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
```

anstsot (approved):
Good work, approving

src/components/Sum.jsx
VladGraund: Thank a lot :)

PR #1630, opened 2023-04-04
added component Sum

Esteban4769:
[DEMO LINK](https://esteban4769.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,17 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+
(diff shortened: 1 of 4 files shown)


src/App.jsx
```diff
@@ -1,14 +1,17 @@
...
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+
+    <Sum a={-5} b={5} />
+
+    <Sum a={10} b={0} />
+
+    <Sum a={0} b={5} />
+
+    <Sum a={0} b={0} />
```
PavloReutskiy, line 15:
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```

0 is the default value, it's extra here and blank line as well

src/components/Sum.jsx
```diff
@@ -1 +1,16 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({ a, b }) => (
```
PavloReutskiy, line 4:
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```

I'm not sure, but maybe with destructuring this syntax will be better than defaultProps

src/App.jsx
SerhiiKirik: agree

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

SerhiiKirik (approved):
Great 🔥

PR #1631, opened 2023-04-04
Solution

yao157:
[DEMO LINK](https://yao157.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,16 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = (props) => {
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,16 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = (props) => {
+  const { a, b } = props;
```
VitaliyBondarenko1982, line 4:
```suggestion
export const Sum = ({ a, b }) => {
```

VitaliyBondarenko1982 (approved):

PR #1636, opened 2023-04-06
sum component added

DenysKrysanov:
[DEMO LINK](https://DenysKrysanov.github.io/react_sum/)


asmd564 (approved):

src/components/Sum.jsx
```diff
@@ -1 +1,13 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum({ a = 0, b = 0 }) {
```
SerhiiKirik, line 4: also, you can use the arrow function expression for function when you intermediate return a value

SerhiiKirik (approved):
Great 🔥

PR #1639, opened 2023-04-06
add solution

Pozdnya:
[DEMO LINK](https://Pozdnya.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,18 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
+    {/* <p>Sum of 2 and 3 is 5</p>
     <p>Sum of -5 and 5 is 0</p>
     <p>Sum of 10 and 0 is 10</p>
     <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <p>Sum of 0 and 0 is 0</p> */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} />
+    <Sum b={5} />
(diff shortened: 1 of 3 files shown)


AnyaLyaLya (approved):
GJ!

src/App.jsx
```diff
@@ -1,14 +1,18 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
+    {/* <p>Sum of 2 and 3 is 5</p>
```
SerhiiKirik, line 7: Remove all comments

src/components/Sum.jsx
```diff
@@ -1 +1,22 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum({ a, b }) {
+  return (
+    <>
```
SerhiiKirik, line 6: remove unnecessary fragment
Also, answer me please, why we need a fragment?

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

src/components/Sum.jsx
Pozdnya: Fragment use for return few tags

SerhiiKirik: Great)
So, you don't have a few tags here, yeah?

src/components/Sum.jsx
```diff
@@ -1 +1,20 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum({ a, b }) {
```
SerhiiKirik, line 4: Also, you can use arrow function expression for functions, where your intermediately return a value

SerhiiKirik (approved):
Well done 🚀

src/components/Sum.jsx
Pozdnya: I just return one tag, so don't need a wrapper tag

PR #1644, opened 2023-04-07
Sum is working

Zad0rnik:
[DEMO LINK](https://Zad0rnik.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,15 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
+
+Sum.propTypes = {
+  a: PropTypes.number.isRequired,
+  b: PropTypes.number.isRequired,
+};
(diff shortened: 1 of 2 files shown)


Sbavdik (approved):
Good job!

src/components/Sum.jsx
```diff
@@ -1 +1,15 @@
...
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
+
+Sum.propTypes = {
+  a: PropTypes.number.isRequired,
+  b: PropTypes.number.isRequired,
```
jstmpelowycz, line 14: if `a` or `b` may not be passed, how they can be required? 🤔

jstmpelowycz (changes requested):

src/components/Sum.jsx
```diff
@@ -1 +1,20 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum({ a, b }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
```
SerhiiKirik, line 10: also, if you return a value immediately - you can use an arrow function 
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
)
```

SerhiiKirik (approved):
Great 🔥

PR #1647, opened 2023-04-07
create sum app

VenomMB:


makandriival (approved):
GJ!

src/components/Sum.jsx
```diff
@@ -1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
```
SerhiiKirik, line 9: also, if you return a value immediately - you can use an arrow function 
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
)
```

SerhiiKirik (approved):
Great 🔥

PR #1650, opened 2023-04-08
create react sum solution

Vlad-Lischenovskiy:
[DEMO LINK](https://Vlad-Lischenovskiy.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,13 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
BondarenkoDmitriy, line 4: you don't use proptypes, your var must have types use like this:
Sum.propTypes = {
  a: PropTypes.number,
  b: PropTypes.number,
};

BondarenkoDmitriy (changes requested):
change some comments and that all))

src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
jstmpelowycz, line 11: re-read _README.md_ one more time 🙂 — 0 values should not be passed

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1651, opened 2023-04-08
solution

makandriival:
[DEMO LINK](https://makandriival.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,4 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum
+  = ({ a = 0, b = 0 }) => <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,14 @@
 import React from 'react';
 import './App.css';
 
+import { Sum } from './components/Sum';
+
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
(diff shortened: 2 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,4 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum
+  = ({ a = 0, b = 0 }) => <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
```
Hamuud, line 4: Instead of default values = 0, propTypes should be used:
`Sum.propTypes = {
a: PropTypes.number,
b: PropTypes.number,
};`

Hamuud (changes requested):
You have taken the first step towards achieving your dreams. Your project is the beginning of something big, so don't stop there and move on!

src/components/Sum.jsx
makandriival: explain pls

Hamuud:
```suggestion
export function Sum({ a, b }) {
  return (
    <p>
      {`Sum of ${a} and ${b} is ${a + b}`}
    </p>
  );
}

Sum.defaultProps = {
  a: 0,
  b: 0,
};
```

makandriival: why do i have to use it? is there something in the checklist or what?

SerhiiKirik, line 4:
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
);
```

SerhiiKirik (approved):
Well done 🚀

PR #1657, opened 2023-04-08
implement sum function

iHateAntichrist:
[DEMO LINK](https://iHateAntichrist.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  );
+}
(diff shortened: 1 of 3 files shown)


vksazonov (approved):
I'm not sure, but maybe for a better understanding of the purpose of defaultProps and propTypes, it's better to use them in your solution.

src/components/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+  );
+}
```
SerhiiKirik, line 7: also, if you return intermediate a value - you can use arrow function 
```suggestion
export cost Sum = ({ a = 0, b = 0 }) => (
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
)
```

SerhiiKirik (approved):
Great 🔥

PR #1659, opened 2023-04-09
made a solution according to the task

Olekk17:
[DEMO LINK](https://olekk17.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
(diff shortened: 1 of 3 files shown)


Rage1502 (approved):
GJ!

src/components/Sum.jsx
```diff
@@ -1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
```
SerhiiKirik, line 9: also, if you return intermediate a value - you can use arrow function 
```suggestion
export cost Sum = ({ a = 0, b = 0 }) => (
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
)
```

SerhiiKirik (approved):
Great 🔥

PR #1671, opened 2023-04-09
solution

Stambuliak:


src/App.jsx
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
 import './App.css';
 
+import { Sum } from './components/Sum';
```
SerhiiKirik, line 4: remove unnecessary line
```suggestion
import { Sum } from './components/Sum';
```

SerhiiKirik (approved):
Well done 🔥

PR #1672, opened 2023-04-09
Implemented Sum function and exported it

vinnykalisa:


kukla1989 (approved):
good job

src/components/Sum.jsx
```diff
@@ -1 +1,21 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Sum = ({
+  a,
+  b,
+}) => (
```
SerhiiKirik, line 7: Wrap if you have more than 2 params or line more that 80
```suggestion
export const Sum = ({ a, b }) => (
```

SerhiiKirik (approved):
Great 🔥

PR #1680, opened 2023-04-10
solve react_sum

kukla1989:
[DEMO LINK](https://kukla1989.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,14 @@
 import React from 'react';
+
 import './App.css';
+import { Sum } from './components';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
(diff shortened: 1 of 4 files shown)


src/App.jsx
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
+
```
SerhiiKirik, line 2: remove unnecessary line
```suggestion
```

src/components/Sum.jsx
```diff
@@ -1 +1,20 @@
...
+import React from 'react';
+import Proptypes from 'prop-types';
+
+export function Sum({ a, b }) {
+  return (
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
```
SerhiiKirik, line 10: also, if you return a value immediately - you can use an arrow function 
```suggestion
export const Sum = ({ a, b }) => (
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
)
```

SerhiiKirik (changes requested):
Well done 🔥 
But, let's deploy your solution 🧑‍💻 
image

src/components/Sum.jsx
SerhiiKirik: Just for your information

src/App.jsx
kukla1989: its line to divide import from third part library and my code

SerhiiKirik (approved):
Great 🔥

PR #1694, opened 2023-04-21
solution

Dka888:
[DEMO LINK](https://Dka888.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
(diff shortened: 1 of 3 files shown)


danielpopek94 (approved):

src/components/Sum.jsx
```diff
@@ -1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
```
d-wasilewski, line 3: Use a named export for the Sum component using the arrow function like: `export const Sum = ...` instead of `export function Sum(...)`.

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1698, opened 2023-04-24
solution

AnastassiaH:
[DEMO LINK](https://anastassiah.github.io/react_sum/)


src/components/Sum.jsx
```diff
@@ -1 +1,21 @@
-// export function Sum() {}
+import React from 'react';
+
+function Sum({ a = 0, b = 0 }) {
```
Anastasiia-Svintsova, line 3: export this component directly to make tests work correctly. And it's better to use one way to declare components in the project (function expression or function declaration). It's different from the App component now
```suggestion
export const Sum = ({ a = 0, b = 0 }) => {
```

src/components/Sum.jsx
```diff
@@ -1 +1,21 @@
...
+      {' '}
+      {a}
+      {' '}
+      and
+      {' '}
+      {b}
+      {' '}
+      is
+      {' '}
+      {+a + +b}
```
Anastasiia-Svintsova, line 16: Use interpolation instead

Anastasiia-Svintsova (changes requested):

src/components/Sum.jsx
AnastassiaH: Це автосейф так робить(

src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a="2" b="3" />
+    <Sum a="-5" b="5" />
+    <Sum a="10" />
+    <Sum b="5" />
```
IvanFesenko, line 10: Send a number instead of string
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5} />
```

src/components/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${+a + +b}`}</p>
```
IvanFesenko, line 4:
```suggestion
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```

IvanFesenko (changes requested):

IvanFesenko (approved):

PR #1701, opened 2023-04-26
First solution

marshalek37:
[DEMO LINK](https://marshalek37.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,14 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a, b }) {
+  return (
+    <p>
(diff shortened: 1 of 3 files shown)


YBM3 (approved):

src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
...
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  );
+}
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
```
d-wasilewski, line 14: interesting approach, but you can set the default variables as you pass the props with `a = 0`

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1703, opened 2023-04-26
commit message

EvgenyNikitinEvgeny:
https://EvgenyNikitinEvgeny.github.io/react_sum/

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,13 @@
 import React from 'react';
+import { Sum } from './components/Sum';
 import './App.css';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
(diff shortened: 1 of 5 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,17 @@
...
+}) => {
+  (
+    <p>
+      Sum of
+      {a}
+      and
+      {b}
+      is
+      {a + b}
+    </p>
```
IvanFesenko, line 15:
```suggestion
    <p>
      `Sum of ${a} and ${b} is ${a + b}`
    </p>
```

IvanFesenko (changes requested):
Your demo link isn't working, you should fix it. Aslo check my comment

src/App.jsx
```diff
@@ -1,14 +1,15 @@
 import React from 'react';
+import { Sum } from './components/Sum';
 import './App.css';
 
 export const App = () => (
```
Volodymyr-Melnyk-Fron10der, line 5: use only one style to export components
```suggestion
const App = () => (
```

Volodymyr-Melnyk-Fron10der (approved):
good👍

PR #1713, opened 2023-04-27
Solution

Vaskovics:
<a href='https://vaskovics.github.io/react_sum/'>Demo</a>

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,13 +1,14 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
     {/* Replace paragraphs with Sum componets */}
     {/* And remove commented lines :) */}
   </>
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,3 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (<p>{`Sum of ${a} and ${b} is ${a + b}`}</p>);
```
anstsot, line 3: please do not forget about code formatting
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
);
```

anstsot (changes requested):
Looks good from code perspective but your demo link isn't working. could you check please?

src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
loralevitska, line 11:
```suggestion
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```

loralevitska (changes requested):
With the previous review you had the wrong parameters passed, the default value was fine.
"no params at all." - means to simply render the component without parameters, also with the previous points about passing only a certain parameter to the component

src/App.jsx
Vaskovics: ohh..sorry, my bad, I am rushing and not paying attention. Thank you!

qoa11a (approved):

PR #1733, opened 2023-05-04
Solution

TarasV123:
[DEMO LINK](https://TarasV123.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,20 @@
-// export function Sum() {}
+export function Sum() {
+  return (
+    <>
+      <Paragraph a = {2} b = {3} />
+      <Paragraph a = {-5} b = {5} />
+      <Paragraph a = {10}/>
+      <Paragraph b = {5} />
+      <Paragraph />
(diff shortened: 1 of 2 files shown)


tetiana-ivanchenko (changes requested):
1. fix demo link
2. fix linter errors
3. You should rewrite your solution according to the task (instead of call paragraph you should call Sum where .you will pass props)
image

src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a, b }) {
+  return (
+    <>
+      <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
+    </>
```
TarasHoliuk, line 7:
```suggestion
    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
react fragment is redundant if the JSX you return has 1 parent (high-level) tag - FYI

TarasHoliuk (approved):

PR #1743, opened 2023-05-16
react_sum solution

mykola-chernysh:
[DEMO LINK](https://mykola-chernysh.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,15 +1,10 @@
 /* eslint-disable react/jsx-one-expression-per-line */
 import React from 'react';
 
-export function Sum(props) {
+export function Sum({ a = 0, b = 0 }) {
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,15 @@
-// export function Sum() {}
+/* eslint-disable react/jsx-one-expression-per-line */
+import React from 'react';
+
+export function Sum(props) {
```
polosanya, line 4:
```suggestion
export function Sum({ a, b }) {
```
It is better to destructure a and b from props object to make your code a bit cleaner

polosanya (changes requested):
Good job! Let's improve your code a bit)

tetiana-ivanchenko (approved):

PR #1744, opened 2023-05-17
solution

AnastasiiaVorobets:
[DEMO LINK](https://AnastasiiaVorobets.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -3,11 +3,9 @@ import PropTypes from 'prop-types';
 
 export function Sum({ a, b }) {
   return (
-    <>
-      <p>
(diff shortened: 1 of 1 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,22 @@
-// export function Sum() {}
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export function Sum({ a, b }) {
+  return (
+    <>
```
vlad-demchuk, line 6: redundant react fragment

vlad-demchuk (changes requested):

YegorKorenkov (approved):
Nice work!

PR #1746, opened 2023-05-18
add react_sum solution

DidenkoDenis306:
[DEMO LINK](https://DidenkoDenis306.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,23 @@
 import React from 'react';
+import { Sum } from './components/Sum';
 import './App.css';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum
+      a={2}
+      b={3}
+    />
+    <Sum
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => (
+  <>
```
33Orange, line 4: The fragment '<>' can be removed, as it's not necessary when there is only one child element.

src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
...
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  </>
```
33Orange, line 8: The closing fragment '</>' can be removed, as it's not necessary when there is only one child element.

src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
+    <Sum />
```
33Orange, line 21: `[CODE STYLE] - If you have < 3 attribues on a tag and the values are short write the tag in one line (to make it easier to write and read).`
Checklist p1.
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```

33Orange (changes requested):
Well done! 

A few moments with readability.

dinaminahi (approved):
Well done! 🥇

PR #1750, opened 2023-05-22
solved

generatedman:


src/components/Sum.jsx
```diff
@@ -1 +1,11 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => {
```
DidenkoDenis306, line 3: You need to set default values ​​in defaultProps

DidenkoDenis306 (changes requested):
Good job, but there is one moment to fix)

dinaminahi (approved):
Good job! 👍

src/components/Sum.jsx
dinaminahi: I didn't find it in the checklist, so destructuring also can be used here

PR #1759, opened 2023-05-27
solution

IamLoren:
https://iamloren.github.io/react_sum/

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,23 @@
 import React from 'react';
+import { Sum } from './components/Sum';
 import './App.css';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
+    <Sum
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
```
VitaliyBondarenko1982, line 20:
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5}  />
```
no need to write each prop on new line in case when it is one or two

VitaliyBondarenko1982 (approved):

PR #1762, opened 2023-05-30
add task solution

AleksandrTsivilin:
[DEMO LINK](https://alexKherson.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -6,8 +6,8 @@ export const App = () => (
   <>
     <Sum a={2} b={3} />
     <Sum a={-5} b={5} />
-    <Sum a={10} b={0} />
-    <Sum a={0} b={5} />
-    <Sum a={0} b={0} />
+    <Sum a={10} />
+    <Sum b={5} />
+    <Sum />
   </>
 );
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,10 @@
-// export function Sum() {}
+/* eslint-disable react/jsx-one-expression-per-line */
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
```
naziamoff, line 5: let's use arrow syntax
```suggestion
const Sum = ({ a = 0, b = 0 }) => (
```

src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
naziamoff, line 11: do you really need to pass 0s? Seems it will be set as default if the value is not passed

naziamoff (changes requested):

PR #1763, opened 2023-05-31
React sum solution

annperetiatko:
[DEMO LINK]
https://annperetiatko.github.io/react_sum/

Also:
- additional spaces appeared after npm run lint was run;
- linter doesn't allow for multiple expressions to be placed at the same line;

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,18 @@
 // export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    Sum of
+    {' '}
+    {a}
+    {' '}
+    and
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,18 @@
 // export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    Sum of
```
mykhalenych, line 6: you can use template string
```<p>{`Sum of ${a} and ${b} is ${a + b}`}</p>```

mykhalenych (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #1764, opened 2023-05-31
solution

skinnyjazz:
[DEMO LINK](https://skinnyjazz.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => {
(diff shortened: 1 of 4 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => {
+  const content = `Sum of ${a} and ${b} is ${a + b}`;
+
+  return <p>{content}</p>;
+};
```
naziamoff, line 7: I don't think there's a need to create a variable, as with this approach you don't win neither in readability nor in ergonomics

naziamoff (approved):
great job!

PR #1766, opened 2023-05-31
add task solution

KatyaZaremska:
[DEMO LINK](https://KatyaZaremska.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,9 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
+      Sum of {a} and {b} is {a + b}
+    </p>
+  );
+}
```
naziamoff, line 9: I would use an arrow syntax here
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
    <p>
      Sum of {a} and {b} is {a + b}
    </p>
 );
```

PR #1767, opened 2023-05-31
initial commit

44shaggy:
[DEMO LINK](https://44shaggy.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,14 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a, b }) {
+  return (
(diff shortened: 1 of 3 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,14 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a, b }) {
```
Anastasiia-Svintsova, line 3: You can use this syntax to add a default value 
```suggestion
export function Sum({ a = 0, b = 0 }) {
```

Anastasiia-Svintsova (approved):

PR #1774, opened 2023-06-02
solution

daryna-hnidash:
[DEMO LINK](https://daryna-hnidash.github.io/react_sum/)

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => {
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
+};
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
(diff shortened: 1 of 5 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,10 @@
...
+import React from 'react';
+
+export const Sum = ({ a, b }) => {
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
+};
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
```
demetrxx, line 10: This signature only works for class components. Use `{ prop = 10 , prop2 = 20 } `instead

src/components/Sum.jsx
```diff
@@ -1 +1,10 @@
-// export function Sum() {}
+import React from 'react';
+
+export const Sum = ({ a, b }) => {
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>;
+};
```
demetrxx, line 5: Function component should return the markup. to dispaly it.

demetrxx (changes requested):
Please, introduce the approriate changes and re-submit the task.

TarasHoliuk (approved):

PR #1790, opened 2023-06-04
Solution

MaksimKad:
[DEMO LINK](https://MaksimKad.github.io/react_sum/)


dinaminahi (changes requested):
Please, fix demolink, because it is not working
and it can be approved

33Orange (approved):
Well done

PR #1805, opened 2023-06-06
added task solution

holytimy4:
[DEMO LINK](https://holytimy4.github.io/react_sum/)


33Orange (changes requested):
Almost done
Add default value for props 

image

silchyn (changes requested):
image

PR #1809, opened 2023-06-06
done

yaroslav5481:
https://yaroslav5481.github.io/react_sum/

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,12 +1,7 @@
 import React from 'react';
 
-export const Sum = ({ a, b }) => (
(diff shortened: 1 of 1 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,13 @@
 // export function Sum() {}
```
OleziO, line 1: Remove all comments from the code.

OleziO (changes requested):

OleziO (changes requested):

OleziO (approved):

PR #1814, opened 2023-06-08
Solution

HannaVasylieva:
https://HannaVasylieva.github.io/react_sum/

### DIFF
--- a/src/components/Sum.jsx
+++ b/src/components/Sum.jsx
@@ -1,1 +1,19 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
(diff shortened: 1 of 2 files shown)


src/components/Sum.jsx
```diff
@@ -1 +1,19 @@
-// export function Sum() {}
+import React from 'react';
+
+export function Sum({ a = 0, b = 0 }) {
+  return (
+    <p>
```
mykhalenych, line 5:
```suggestion
    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```

mykhalenych (changes requested):

PR #1815, opened 2023-06-08
react_sum

vintwp:


src/App.jsx
```diff
@@ -1,14 +1,14 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
nazarmatsevych, line 12:
```suggestion
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```

No need to pass zeros since you set zero as a default value

src/components/Sum.jsx
```diff
@@ -1 +1,15 @@
...
+  b = 0,
+}) => (
+  <p>
+    Sum of
+    {` ${a} `}
+    and
+    {` ${b} `}
+    is
+    {` ${a + b}`}
+  </p>
```
nazarmatsevych, line 14:
```suggestion
  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```

Use [template string ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

nazarmatsevych (changes requested):
Good work!
Pls check comments to improve your solution

anastasiiavorobiova (approved):
Great job!

PR #1828, opened 2023-06-17
code

liza-strykharchuk:
[DEMO LINK](https://liza-strykharchuk.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,28 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
 export const App = () => (
   <>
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -1,14 +1,28 @@
...
+      b={0}
+    />
+    <Sum
+      a={0}
+      b={5}
+    />
+    <Sum
+      a={0}
+      b={0}
+    />
```
TarasHoliuk, line 26:
```suggestion
    <Sum />
```
FYI: you may not pass `0` value since you set it as default for `a` and `b` props in the `Sum` component

TarasHoliuk (approved):
Added 1 comment just FYI (no need to re-request review - already approved)

PR #1846, opened 2023-06-26
React Sum

ilona-romaniuk:
https://ilona-romaniuk.github.io/react_sum/

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,27 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,27 @@
...
+    />
+
+    <Sum
+      a={-5}
+      b={5}
+    />
+
+    <Sum
+      a={10}
+    />
```
witflash, line 19: You can place a component with a single prop in a single line:
```
<Sum a={10} />
```
It's not critical but looks better.

PR #1855, opened 2023-06-27
Solution

Viktoriia6666:
[DEMO LINK](https://Viktoriia6666.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,24 @@
+
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -1,14 +1,24 @@
...
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
+    <Sum />
```
loralevitska, line 22: if there are enough attributes to write everything in one line and not reach 80 characters in a line, it is better to write in one line so as not to increase the number of lines where it is not needed

```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```

loralevitska (approved):
Well done)

PR #1893, opened 2023-06-28
add task solution React_sum

TamaraShukalo:
[DEMO LINK](https://TamaraShukalo.github.io/react_sum/)

### DIFF
--- a/src/components/Sum/Sum.jsx
+++ b/src/components/Sum/Sum.jsx
@@ -1,5 +1,3 @@
 export const Sum = ({ a = 0, b = 0 }) => (
-  <>
-    <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
-  </>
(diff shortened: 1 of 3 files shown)


outofs (changes requested):
GW. But need some changes.

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export const Sum = () => ();
+export const Sum = ({ a = 0, b = 0 }) => (
+  <>
```
outofs, line 2: You don't need this `<></>` wrapper, because your component has only one `<p><p/>` tag. Maybe that's a reason why  some tests are failed

outofs (changes requested):

tetiana-ivanchenko (approved):

PR #1901, opened 2023-06-29
add task solution

EvgeniyChumak:
https://EvgeniyChumak.github.io/react_sum/

### DIFF
--- a/src/components/Sum/Sum.jsx
+++ b/src/components/Sum/Sum.jsx
@@ -1,1 +1,15 @@
-// export const Sum = () => ();
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    Sum of
+    {' '}
+    { a }
+    {' '}
+    and
+    {' '}
+    { b }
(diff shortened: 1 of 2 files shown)


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,15 @@
...
+    { a }
+    {' '}
+    and
+    {' '}
+    { b }
+    {' '}
+    is
+    {' '}
+    {a + b }
+  </p>
```
tetiana-ivanchenko, line 14: image

tetiana-ivanchenko (changes requested):

src/components/Sum/Sum.jsx
EvgeniyChumak: The problem is that when I run 'npm run lint' or 'git commit - m' it automatically change my code into solution that u consider as wrong.

YegorKorenkov (approved):
GJ!

PR #1911, opened 2023-06-30
finished

EvhenIvchenko:
[DEMO LINK](https://EvhenIvchenko.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,27 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -1,14 +1,27 @@
...
+
+    <Sum
+      a={10}
+    />
+
+    <Sum
+      b={5}
+    />
+
+    <Sum />
```
loralevitska, line 25: if the number of characters allows you to write a string in one line, then it is better to do so in order not to increase the number of lines for easy reading of the code
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```

loralevitska (approved):
Well done)

PR #1912, opened 2023-07-01
Homework

Skrudj:


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,6 @@
-// export const Sum = () => ();
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
SerhiiKirik, line 4: wrap if you have > 2 params or length > 80
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```

SerhiiKirik (approved):
Nice 🔥

PR #1913, opened 2023-07-01
Solution

Vasyl-Zhyliakov:
https://Vasyl-Zhyliakov.github.io/react_sum/

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,23 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
+    <Sum />
```
loralevitska, line 21: if the length of the characters in the line allows you to write the expression in one line, then it is better to do so in order not to increase the amount of code
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5}  b={5} />
    <Sum a={10} />
    <Sum b={5}  />
    <Sum />
```

loralevitska (approved):
Well done)

src/App.jsx
Vasyl-Zhyliakov: thanks)

PR #1934, opened 2023-07-12
solution

vshepel:
[DEMO LINK](https://vshepel.github.io/react_sum/)


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export const Sum = () => ();
+export const Sum = ({ a = 0, b = 0 }) => (
+  <>
```
mykhalenych, line 2: redundant fragment

mykhalenych (changes requested):

PR #1935, opened 2023-07-12
add task solution

Vitalii-Maistrenko:
https://Vitalii-Maistrenko.github.io/react_sum/


lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

agniya-i (approved):

PR #1936, opened 2023-07-12
Solution

Nataliia95254:
https://Nataliia95254.github.io/react_sum/

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,28 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
 export const App = () => (
   <>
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -1,14 +1,28 @@
...
+      b="0"
+    />
+    <Sum
+      a="0"
+      b="5"
+    />
+    <Sum
+      a="0"
+      b="0"
+    />
```
TarasHoliuk, line 26:
```suggestion
    <Sum />
```
you may not pass props with values the same as default value in the component - FYI

TarasHoliuk (approved):
Added 1 comment just FYI - approved :heavy_check_mark:

PR #1937, opened 2023-07-13
solution

ShamievDavid:
[DEMO LINK](https://ShamievDavid.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,32 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
 export const App = () => (
   <>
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,32 @@
...
+      b={0}
+    />
+
+    <Sum
+      a={0}
+      b={5}
+    />
+
+    <Sum
+      a={0}
```
OleziO, line 28: Make it without props. Use default values.

OleziO (changes requested):

src/App.jsx
ShamievDavid: sure

mykhalenych (approved):

PR #1938, opened 2023-07-13
react sum is done

AllaNebozhenko:
[DEMO LINK](https://AllaNebozhenko.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,23 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum
+      a={2}
+      b={3}
+    />
```
polosanya, line 10:
```suggestion
    <Sum a={2} b={3} />
```
Fix all cases
image

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,10 @@
-// export const Sum = () => ();
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
polosanya, line 4:
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```
Here it also can be in one line if the amount of props is less than 3

polosanya (changes requested):
Almost perfect!

PR #1954, opened 2023-07-19
Home work is done

georgiy-nezbritskiy:


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,9 @@
 // export const Sum = () => ();
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
polosanya, line 5:
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```
You can place it in one line, as there are < 3 props and the line is not too long

polosanya (changes requested):
Well done! Please deploy the solution and add a demo link to the PR description
image

polosanya (approved):
🔥

PR #1971, opened 2023-07-27
commit

OleksChernikov:
https://OleksChernikov.github.io/react_sum/

### DIFF
--- a/src/components/Sum/Sum.jsx
+++ b/src/components/Sum/Sum.jsx
@@ -1,1 +1,17 @@
-// export const Sum = () => ();
+import React from 'react';
+
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    Sum of
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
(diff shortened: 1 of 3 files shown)


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,17 @@
...
+    {' '}
+    {a}
+    {' '}
+    and
+    {' '}
+    {b}
+    {' '}
+    is
+    {' '}
+    {a + b}
```
Anastasiia-Svintsova, line 15: It's hard to read now. Consider using [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for better readability

Anastasiia-Svintsova (changes requested):
Almost done!

src/components/Sum/Sum.jsx
OleksChernikov: corrected

lerastarynets (approved):
You're doing amazing, keep it up!

PR #1991, opened 2023-08-02
solution

HannieA:
https://HannieA.github.io/react_sum/

### DIFF
--- a/src/components/Sum/Sum.jsx
+++ b/src/components/Sum/Sum.jsx
@@ -1,1 +1,13 @@
-// export const Sum = () => ();
+/* eslint-disable react/jsx-one-expression-per-line */
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => {
+  const sum = a + b;
+
+  return (
+    <p>Sum of {a} and {b} is {sum}</p>
+  );
+};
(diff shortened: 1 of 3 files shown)


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,13 @@
...
+import React from 'react';
+
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => {
+  const sum = a + b;
+
+  return (
+    <p>Sum of {a} and {b} is {sum}</p>
```
lerastarynets, line 11: The rendered string could be improved for better readability. Consider using template literals to interpolate the variables in the string.

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,13 @@
-// export const Sum = () => ();
+/* eslint-disable react/jsx-one-expression-per-line */
```
lerastarynets, line 1: It is generally a bad practice to disable ESLint rules unless absolutely necessary. You should consider re-enabling the react/jsx-one-expression-per-line rule and adjust your code to comply with it.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

PR #1992, opened 2023-08-03
task solution 02.08.23

AnVBondar:


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum
+      a={2}
+      b={3}
+    />
```
volodymyr-soltys97, line 10: Need to check and fix it everywhere
```suggestion
    <Sum a={2} b={3} />
```

volodymyr-soltys97 (changes requested):

src/App.jsx
AnVBondar: Fixed.

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,10 @@
-// export const Sum = () => ();
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
+  <>
```
anastasiiavorobiova, line 5:
```suggestion

```

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,10 @@
...
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  </>
```
anastasiiavorobiova, line 9:
```suggestion

```

anastasiiavorobiova: You are using just one element, so there is no need to add fragments here

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/components/Sum/Sum.jsx
AnVBondar: Clear. Thanks.

anastasiiavorobiova (approved):
Awesome!

PR #2010, opened 2023-08-08
sum-app

Orchibald:
[DEMO LINK](https://orchibald.github.io/react_sum/)


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
+    <Sum />
```
VitaliyBondarenko1982, line 21:
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```
no need to make each prop in new line in this case (just two props maximum and they are rather shorts)

VitaliyBondarenko1982 (approved):
Looks good

PR #2023, opened 2023-08-14
Solution

NataliiaHen:
[DEMO LINK](https://NataliiaHen.github.io/react_sum/

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,23 @@
 import React from 'react';
 import './App.css';
+import { Sum } from './components/Sum/Sum';
 
 export const App = () => (
-  <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
+    <Sum />
```
IvanFesenko, line 21: Props same as html atributes code style - if there are 2 or less use single line
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,10 @@
...
+export const Sum = ({ a, b }) => (
+  <p>
+    {`Sum of ${a} and ${b} is ${a + b}`}
+  </p>
+);
+
+Sum.defaultProps = {
+  a: 0,
+  b: 0,
+};
```
IvanFesenko, line 10: Also you can set default value this way
```suggestion
export const Sum = ({ a= 0, b = 0 }) => (
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
);
```

IvanFesenko (approved):
Good job

PR #2026, opened 2023-08-16
React_Sum

levchenko-dmytro:
[DEMO LINK](https://levchenko-dmytro.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,24 @@
 import React from 'react';
 import './App.css';
 
+import { Sum } from './components/Sum/Sum';
+
 export const App = () => (
   <>
-    <p>Sum of 2 and 3 is 5</p>
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum
+      a={2}
+      b={3}
+    />
+    <Sum
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,24 @@
...
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum
+      a={2}
+      b={3}
+    />
```
anastasiiavorobiova, line 11:
```suggestion
    <Sum  a={2}  b={3} />
```
There are just 2 props, so there is no need to separate them into multiple lines

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,12 @@
-// export const Sum = () => ();
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    <>
```
anastasiiavorobiova, line 3: You don't need to use Fragment here. 
The component returns everything wrapped inside the 'p' tag. 
It doesn't return multiple elements.

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,12 @@
...
+      <span>Sum of </span>
+      {a}
+      <span> and </span>
+      {b}
+      <span> is </span>
+      {a + b}
```
anastasiiavorobiova, line 9: Consider using a template string instead of spans. You can use a span wrapper to add some styling for the text, but there is no additional styling in this task.

anastasiiavorobiova (changes requested):
Great start! Keep doing

mykhalenych (approved):

anastasiiavorobiova (approved):
Amazing!

PR #2027, opened 2023-08-16
solution

Moblat890:


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
-    <p>Sum of -5 and 5 is 0</p>
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum
+      a={2}
+      b={3}
+    />
```
TarasHoliuk, line 10:
```suggestion
    <Sum a={2} b={3}  />
```
it will be easier to read
write each prop on the new line if there are more than 3 props (the same as with HTML attributes)

TarasHoliuk (changes requested):

PR #2055, opened 2023-08-21
add task solution

valet2102:


Moroz-Dmytro (changes requested):
It works well, but if tests are failed, you need to fix them
Pay attention to tests(check how your props should be named)

PR #2060, opened 2023-08-21
Update README.md

matuschak90565:


anastasiiavorobiova (changes requested):
Unfortunately couldn't see the solution and the demo link here

PR #2082, opened 2023-08-22
add task solution

edgergard:
[DEMO LINK](https://edgergard.github.io/react_sum/)


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,8 @@
-// export const Sum = () => ();
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
tiserett, line 4: better put everything on the same line, it's not that long ;)

tiserett (approved):

PR #2083, opened 2023-08-22
commit 1

denys2:
https://denys2.github.io/react_sum/


witflash (changes requested):
Hi!
Your DEMO link looks good.
But I don't see your working files in your PR. I think you forgot to commit and push all your changes.
Check it, please.

witflash (changes requested):
Now there is `App` file. But what about `Sum` component? )
image
Add all your source files with the command `git add src`

sTorba24 (approved):

PR #2085, opened 2023-08-22
solution

AndrMar1939:
[DEMO LINK](https://AndrMar1939.github.io/react_sum/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,14 +1,23 @@
 import React from 'react';
+import { Sum } from './components/Sum/Sum';
 import './App.css';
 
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
+    <Sum
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
```
tiserett, line 20: let's put every property on component on the same line with it, they're not that long ;)

tiserett (approved):

PR #2086, opened 2023-08-22
add task solution

NazarVotkanych:
Demo: https://NazarVotkanych.github.io/react_sum/


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,8 @@
-// export const Sum = () => ();
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
tiserett, line 4: let's put everything on the same line, since it's kinda short

tiserett (approved):

PR #2095, opened 2023-08-22
sol

VolodimirKurhan:


src/App.jsx
```diff
@@ -1,14 +1,23 @@
...
+      a={-5}
+      b={5}
+    />
+    <Sum
+      a={10}
+    />
+    <Sum
+      b={5}
+    />
+    <Sum />
```
VitaliyBondarenko1982, line 21:
```suggestion
    <Sum a={2} b={3} />
    <Sum a={-5} b={5} />
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```
it is redundant to make each prop on a new line in case if them max two and they too short

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,8 @@
-// export const Sum = () => ();
+export const Sum = ({
+  a = 0,
+  b = 0,
+}) => (
```
VitaliyBondarenko1982, line 4:
```suggestion
export const Sum = ({ a = 0, b = 0 }) => (
```
in this case not required to make each prop on new line.

VitaliyBondarenko1982 (approved):
Good job.
Please pay attention on comment which I left for future tasks

PR #2097, opened 2023-08-22
add task solution

k1NAs:
[DEMO LINK](https://k1NAs.github.io/react_sum/)


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,7 @@
-// export const Sum = () => ();
+export const Sum = ({ a = 0, b = 0 }) => (
+  <>
+    <p>
+      {`Sum of ${a} and ${b} is ${a + b}`}
+    </p>
+  </>
```
tiserett, line 6: remove fragment, it's redundant here

PR #2122, opened 2023-08-24
react_sum

annak1104:


src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
olegpaunov, line 11: You can write only one parametr
And at the end you can write components without anything

olegpaunov (approved):

src/App.jsx
Moroz-Dmytro, line 11: By task description here you should pass only 1 prop to your component, or even use a component without passing props. You don't need to pass prop `a=0` for example, as it is value by default

Moroz-Dmytro, line 11:
```suggestion
    <Sum a={10} />
    <Sum b={5} />
    <Sum />
```

Moroz-Dmytro (changes requested):
Small comment

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,3 @@
-// export const Sum = () => ();
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>{`Sum of ${a} and ${b} is ${a + b}`}</p>
```
Moroz-Dmytro, line 2:
```suggestion
  <p>
    {`Sum of ${a} and ${b} is ${a + b}`}
  </p>
```

Moroz-Dmytro (approved):
LGTM 👍

PR #2125, opened 2023-08-24
sum component

olegobiukh:


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,15 @@
-// export const Sum = () => ();
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    Sum of
+    {' '}
```
VitaliyBodnar, line 4: it's a bad practice to add spaces this way, it's better to use margins or paddings for that 
check other places

VitaliyBodnar (changes requested):

PR #2153, opened 2023-09-06
Develop

MarinaOlenchenko:
[DEMO LINK](https://MarinaOlenchenko.github.io/react_sum/)

### DIFF
--- a/src/components/Sum/Sum.jsx
+++ b/src/components/Sum/Sum.jsx
@@ -1,1 +1,10 @@
 // export const Sum = () => ();
+import React from "react";
+
+export const Sum = ({ a = 0, b = 0 }) => {
+  const sum = a + b;
+
+  return <p>
+    Sum of {a} and {b} is {sum}
+    </p>;
+}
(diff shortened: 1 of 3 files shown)


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,20 @@
...
+    <p>
+      Sum of
+      {' '}
+      {a}
+      {' '}
+      and
+      {b}
+      {' '}
+      is
+      {sum}
```
etojeDenys, line 17: we can rewrite it in one line using [Template literals (Template strings)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

etojeDenys (changes requested):

src/components/Sum/Sum.jsx
```diff
@@ -1 +1,12 @@
 // export const Sum = () => ();
```
lerastarynets, line 1: remove comments

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #2158, opened 2023-09-11
Solution

KarolinaWiktoria:
[DEMO](https://KarolinaWiktoria.github.io/react_sum/)


src/components/Sum/Sum.jsx
```diff
@@ -1 +1,5 @@
-// export const Sum = () => ();
+export const Sum = ({ a = 0, b = 0 }) => (
+  <p>
+    {`Sum of ${a} and ${b} is ${a + b}`}
```
Centmsn, line 3: While this works, remember that you do not have to open (and close) a string while in JSX, so the code below works exactly the same way but is easier to read:
```javascript
<p>
  Sum of {a} and {b} is {a + b}
</p>
```

src/App.jsx
```diff
@@ -1,14 +1,13 @@
...
-    <p>Sum of 10 and 0 is 10</p>
-    <p>Sum of 0 and 5 is 5</p>
-    <p>Sum of 0 and 0 is 0</p>
-    {/* Replace paragraphs with Sum componets */}
-    {/* And remove commented lines :) */}
+    <Sum a={2} b={3} />
+    <Sum a={-5} b={5} />
+    <Sum a={10} b={0} />
+    <Sum a={0} b={5} />
+    <Sum a={0} b={0} />
```
Centmsn, line 11: Props can be completely skipped like so:
```jsx
<Sum a={10} />
<Sum b={5} />
<Sum />
```
If `a` or `b` is not provided then the default value (`0` in this cause is used).

Centmsn (commented):
Good job! 😃

I left you a few comments; nothing serious, just small improvements to the code. However, they are worth remembering while writing React code.

src/components/Sum/Sum.jsx
KarolinaWiktoria: It doesn't work with lint, so I leave it as before.

plodnikate (approved):
good job

PR #4013, opened 2025-12-11
add task solution

Arsenii34:
https://Arsenii34.github.io/react_sum/


Arsenii34:
https://Arsenii34.github.io/react_sum/

PR #4175, opened 2026-03-29
add task solution

miftaha:
[DEMO LINK](https://miftaha.github.io/react_sum/)


miftaha:
Done — Sum component implemented (named export); App renders required Sum instances; defaults handled; ready for review.

PR #4205, opened 2026-04-15
Soluiton

protocolo23:
[DEMO LINK](https://protocolo23.github.io/react_sum/)


raulriato (approved):
As the mentor AI suggested, you're not rendering the `Sum.jsx` the way it's supposed to. But, since it's not a major problem, I'll approve the task.

protocolo23:
> As the mentor AI suggested, you're not rendering the `Sum.jsx` the way it's supposed to. But, since it's not a major problem, I'll approve the task.

Thanks for the feedback. I’ll make sure to fix the issue with the Sum.jsx as Like requested.

